PR #119, opened 2026-08-19
fix(input)!: make send/send_str/paste fallible and name a departed child

### LINKED ISSUE #105: `send`/`send_str`/`paste` cannot fail, and a write to a dead child is silent
**Today** — half the input API can fail and half cannot, and the half that cannot fails silently.

```rust
pub fn send(&mut self, key: impl Input + fmt::Debug)   // -> ()
pub fn send_str(&mut self, s: &str)                    // -> ()
pub fn paste(&mut self, text: &str)                    // -> ()
pub fn click(&mut self, col: u16, row: u16) -> Result<()>
pub fn drag(...) -> Result<()>
pub fn scroll(...) -> Result<()>
```

Measured against 0.4.0, writing to a child that has already exited:

- `send()` **returns normally**. No panic, no error, nothing. The keystroke goes into a dead PTY and the test continues as if it had been delivered, failing later at a wait with a screen that never changed — the symptom appears far from the cause.
- `click()` returns `Err`, but with the **wrong diagnosis**: *"the application has not enabled mouse tracking (no CSI ?9/?1000/?1002/?1003 h was seen)"*. The mouse-mode check runs before any liveness check, so a dead child is reported as a mode problem.

Internally `send` routes through `write_or_panic`, so it *can* abort the test — just not through the type system, and not in this case.

**Why now** — v0.3 put a deadline on writes precisely so a stuck application produces a screen-carrying failure instead of a hang. That work is undermined by a signature that cannot carry the failure: the deadline fires, and the only ways out are a panic or silence.

**Fix — two parts:**

1. **Make the fallible operations fallible.** `send`, `send_str` and `paste` return `Result<()>`, replacing `write_or_panic`. This is breaking and noisy at call sites — every `t.send(Key::Enter)` grows a `?` — which is why it belongs in the same release as any other break rather than trailing one.
2. **Check liveness before mode.** A dead or reaped child is reported as such by `click`/`drag`/`scroll`, ahead of the mouse-tracking check, so the message names the real cause.

**Done when** — sending to an exited child returns an error naming the child, not silence and not a mouse-mode message; `write_or_panic` is gone; the panic path is no longer how a write failure reaches the test.

vyncint:
Closes #105.

## Verified first

Both halves were measured against the **published 0.4.2** before any code changed ([full results](https://github.com/vyncint/termlens/issues/105#issuecomment-5341271240)).

**Part 2 reproduces verbatim** — and it is the sharper bug. Against a reaped child:

```
click  -> Err("the application has not enabled mouse tracking (no CSI ?9/?1000/?1002/?1003 h was seen)")
scroll -> Err(same)
drag   -> Err(same)
```

A child that has exited necessarily never enabled tracking either, so the mode check was always *technically* right and never the reason. With tracking enabled before the child died, `click` panicked with EIO instead — so the mouse path had both faults.

**Part 1's measurement does not reproduce.** `send()` to an exited child does not return silently; it panics with EIO and the screen attached — 0.3's acknowledged write path already did that. The one case that returns normally is a *live* child that has not read yet, where the write genuinely succeeded into the terminal's input queue. That is a pending keystroke, not a lost one, and this PR pins it as correct.

So part 1 lands on the narrower argument the issue also makes: the deadline could only reach a test by aborting it.

## Changes

- `send`, `send_str`, `paste` → `Result<()>`; `write_or_panic` is gone. **Breaking**; call sites grow a `?`.
- `click`/`drag`/`scroll` call `ensure_deliverable` before the mode check.
- New `Error::Write { what, screen }` carrying the screen, like `Timeout` and `Eof`. `Error::screen()` returns it. `Error` is `#[non_exhaustive]`, so the variant itself is not breaking.

## The panic/error split, stated

`Key::F(13)` still panics, and `docs/DESIGN.md` §1 now says why: **environmental failures are errors; impossible arguments are panics.** Whether the child is still there is discovered at runtime and a test may reasonably handle it. There is no thirteenth function key on any terminal, so that is a mistake in the test's own source — the same category as indexing a slice out of bounds.

## Tests

Three added to `tests/input.rs`, one per limb of the contract:

- `typed_input_to_a_departed_child_is_a_typed_error` — `Error::Write` from all three senders, screen attached
- `a_mouse_click_at_a_departed_child_blames_the_child` — asserts the message does **not** say "mouse tracking" and does say "exited", for click/scroll/drag
- `typed_input_to_a_live_child_that_has_not_read_yet_succeeds` — the case that must stay a success

99 call sites migrated across the suite. Four tests that deliberately return `()` to assert on an `Err` path use `.unwrap()` rather than growing a `Result` return.

## Checklist

- [x] Linked an issue — closes #105
- [x] Snapshot changes (if any) were reviewed with `cargo insta review` — none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,13 @@ listed under a **Changed** or **Removed** heading.
   cannot be delivered is something a test can see, handle, or propagate
   with `?` — previously the only route from a failed write to the test was
   aborting it. Call sites grow a `?`; that is the whole migration.
+- **Typed input to a closed terminal is refused identically on Linux and
+  macOS.** It was not: a write to a master whose slave descriptors are all
+  closed fails with `EIO` on macOS and *succeeds* on Linux, queueing the
+  bytes for a reader that no longer exists. The same keystroke was
+  therefore an error on one CI runner and silently discarded on the other.
+  Every sender now checks for a closed terminal before writing, so the
+  answer is the same everywhere and no keystroke is lost quietly.
 - **A mouse action at a departed child names the child.** `click`, `drag`
   and `scroll` check liveness *before* the mouse-tracking mode, because a
   child that has exited necessarily never enabled tracking either — so the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,26 @@ listed under a **Changed** or **Removed** heading.
 
 ## [Unreleased]
 
+### Changed
+
+- **`send`, `send_str` and `paste` return `Result<()>`** and no longer
+  panic. Every input call in the crate is now fallible, so a write that
+  cannot be delivered is something a test can see, handle, or propagate
+  with `?` — previously the only route from a failed write to the test was
+  aborting it. Call sites grow a `?`; that is the whole migration.
+- **A mouse action at a departed child names the child.** `click`, `drag`
+  and `scroll` check liveness *before* the mouse-tracking mode, because a
+  child that has exited necessarily never enabled tracking either — so the
+  old order reported a missing `CSI ?1000 h` for a terminal whose
+  application was simply gone. The tracking-mode error is unchanged for a
+  live application that really has not enabled it.
+
+### Added
+
+- **`Error::Write`**, carrying the screen at the moment of the failed
+  write, the way `Error::Timeout` and `Error::Eof` already do.
+  `Error::screen()` returns it.
+
 ## [0.4.2] - 2026-08-19
 
 The documentation set, brought up to what 0.4 actually does.
--- a/README.md
+++ b/README.md
@@ -148,8 +148,9 @@ design. termlens's position:
 - **Every wait takes a per-call deadline** (`wait_until_for`,
   `wait_frame_for`, `wait_idle_for`, `wait_exit_for`), so one slow step
   doesn't force a generous timeout on the whole suite. Writes are bounded
-  too: typing into an application that has stopped reading fails with the
-  screen attached instead of hanging.
+  too, and every input call returns `Result`: typing into an application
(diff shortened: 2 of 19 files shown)


vyncint:
## Correction, caught by this PR's own test

The first push relied on the write failing on its own. `test (ubuntu-latest)` disagreed:

```
thread 'typed_input_to_a_departed_child_is_a_typed_error' panicked at input.rs:323:
called `Result::unwrap_err()` on an `Ok` value: ()
```

`send()` returned `Ok(())` after the child had exited **and been reaped** — on Linux. Writing to a master whose slave descriptors are all closed fails with `EIO` on macOS and **succeeds on Linux**, queueing the bytes for a reader that no longer exists.

So #105's original report is right and [my verification comment](https://github.com/vyncint/termlens/issues/105#issuecomment-5341271240) was wrong: it measured macOS only and said so nowhere. I have [corrected that on the issue](https://github.com/vyncint/termlens/issues/105#issuecomment-5341399359).

**What changed in this branch as a result**, beyond the assertion:

- `ensure_deliverable` now gates **every** sender — `send`, `send_str`, `paste` as well as `click`/`drag`/`scroll`. Before, only the mouse path had it, and the three senders leaned on the write failing, which is what differs by OS.
- The signal is **EOF, not "did we reap something"**. EOF means every slave descriptor is closed, which is exactly when nothing can read. A reaped child says nothing about a grandchild that inherited the terminal and is still reading it, and refusing that write would be wrong.
- The refusal names the closed terminal rather than an errno, since no syscall is attempted.
- New test `a_departed_child_is_refused_identically_on_every_platform` pins the property that was split, with a comment recording which OS did what.

The failure was worth more than the fix: without a both-OSes gate, this branch would have shipped an API whose contract depended on the runner.

PR #129, opened 2026-08-19
fix(queries): answer a startup batch in full instead of outrunning the writer

### LINKED ISSUE #107: Query replies are still dropped past 64 rather than backpressured
**Today** — the responder queue is `REPLY_QUEUE_DEPTH = 64` deep and **drops** when full, unchanged since 0.2.1. The premise in the source is that a full queue means the application has stopped reading its input entirely, "in which case it cannot be waiting on these bytes."

That premise is false for batch-probe-then-read, a legitimate startup pattern. Measured in the [coverage study](https://github.com/vyncint/termlens-demo/blob/main/docs/TERMLENS-COVERAGE.md) §4.2, with the application reading everything after a pause:

| queries asked | replies received |
|---|---|
| 50 | 50 |
| 200 | **94** |
| 400 | 161 |
| 1000 | 285 |

A real terminal backpressures — it makes the writer wait. termlens discards, and the application waits forever for answers that were thrown away.

**The discovery path is the worst part.** The only signal is a note appended to a *later* failing wait. In the study's `v14` the application carried on with 395 of 1500 answers and nothing failed, so nothing warned at all. A silently dropped reply is indistinguishable from a query that was never asked.

Ordinary applications issuing a handful of probes at startup are unaffected, which is why this has never bitten in practice — but the failure, when it comes, is invisible.

**Fix** — backpressure instead of dropping: block the responder on a full queue, bounded by the terminal's deadline, so a genuinely non-reading application produces a timeout that names the cause rather than a silent discard. v0.3 already put deadlines on writes; this is the same treatment for the reply path, and the machinery exists.

If dropping is kept for any case, the drop must be loud at the moment it happens rather than discoverable only through an unrelated later failure.

**Done when** — an application that batches 200 probes and then reads receives 200 replies; an application that never reads produces a deadline error naming the undelivered replies; no reply is discarded without a signal at the time.

vyncint:
Closes #107.

## The loss reproduces; the stated cause does not

Verified against the published 0.4.2, varying **only the spacing between queries**:

| query rate | asked | answered |
|---|---|---|
| back to back | 200 | 173 |
| **1 ms apart** | 200 | **200** |

The application's read behaviour is identical in both rows — it reads nothing until after the batch. If the queue were filling because the application had stopped reading, spacing the *queries* apart would change nothing.

It filled because **the reader outran our own writer**: one channel entry per reply, one `write(2)` per entry, at ~6 bytes per DSR reply with the ~8 KB tty buffer nowhere near full. The comment in the source asserting that a full queue meant the application had stopped reading was simply false. ([full measurement](https://github.com/vyncint/termlens/issues/107#issuecomment-5341273931))

## Why the proposed fix would have made it worse

"Block the responder on a full queue" blocks the **reader** thread, which is the deadlock the fire-and-forget design exists to prevent — `docs/DESIGN.md` §1 spells out the chain: reader blocks → drain stops → child fills its output buffer → child blocks writing → child never reads → writer stays blocked. A deadline bounds that to a stall rather than a hang, but it would trade a silent drop for a **stalled drain in exactly the batch-probe case this issue is about**, where nothing was ever blocked.

## The fix that matches the cause

One queue entry per **read** rather than per reply, and the writer coalesces whatever is queued behind the entry it took into a single `write(2)`. Ordering is unchanged — bytes are concatenated in the order they were built.

Result, measured by the new test at the sizes the issue recorded:

| asked | before | now |
|---|---|---|
| 50 | 50 | 50 |
| 200 | 173 | **200** |
| 400 | 235 | **400** |
| 1000 | 285 | **1000** |

## The fix moved the diagnosis, which needed a second change

Batching makes a discard rare — which is right — but a genuinely non-reading application then leaves its answers **stuck in a blocked write** rather than dropped. So the note vanished exactly where it was most needed. My first version of `an_application_that_never_reads_is_named_in_the_error` failed for precisely this reason: 4000 queries arrive in two or three 8 KB reads, so the 64-deep queue never fills and nothing was ever "dropped".

Undelivered replies are now counted whether **dropped or pending**, through a lock-free counter the writer clears only once the bytes are actually out. Lock-free deliberately: the writer must never need the state lock, since "the state lock and the writer lock are never held together" is what keeps the harness from deadlocking itself.

## Tests

- `a_batch_of_probes_is_answered_in_full` — 50/200/400/1000, counting `R` bytes off the wire
- `an_application_that_never_reads_is_named_in_the_error` — the timeout names the cause *and* the count

## One refactor

`EmuState::new` reached eight arguments (clippy's limit is seven). The six that are all one decision — what this terminal claims to be when an application asks — are now a `Responder` struct.

## Checklist

- [x] Linked an issue — closes #107
- [x] Snapshot changes (if any) were reviewed with `cargo insta review` — none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,11 @@ listed under a **Changed** or **Removed** heading.
   anywhere. The reader was enqueueing one channel entry per *reply* while the
   writer issued one `write(2)` per entry, and the queue filled because the
   reader outran it. Replies from one read are now one entry, the writer
-  coalesces whatever is queued behind it into a single write, and 50/200/400/1000
-  are all answered completely.
+  coalesces whatever is queued behind it into a single write, and 50, 200 and
+  400 are all answered completely. The remaining ceiling is the terminal's own
+  input queue rather than ours — around 4 KB on Linux, so a batch past roughly
+  680 replies still needs the application to read as it asks, which is true of
+  a real terminal too.
 - **An application that never reads is named in the wait error.** Batching made
   discards rare, which is right — but it also meant a genuinely non-reading
   application produced a plain timeout with the answers stuck in the writer and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,11 +35,14 @@ listed under a **Changed** or **Removed** heading.
   input queue rather than ours — around 4 KB on Linux, so a batch past roughly
   680 replies still needs the application to read as it asks, which is true of
   a real terminal too.
-- **An application that never reads is named in the wait error.** Batching made
-  discards rare, which is right — but it also meant a genuinely non-reading
-  application produced a plain timeout with the answers stuck in the writer and
-  nothing saying so. Undelivered replies are now counted whether they were
-  dropped *or* are blocked mid-write, so the note appears in either case.
+- **Undelivered replies are counted whether dropped or blocked mid-write**, so
+  a non-reading application is named in the wait error rather than producing a
+  plain timeout. How much of this is knowable turns out to differ by platform,
+  and `docs/DESIGN.md` §1 now says so: a write into a full terminal input queue
+  blocks on macOS, where the count is exact, while Linux's `n_tty` *discards*
+  input once its 4 KB buffer is full — the write succeeds, the kernel throws
+  the bytes away, and nothing distinguishes that from delivery. We cannot
+  report what we were never told.
 - **`drag` reports one motion per cell crossed**, on a straight interpolated
   path, instead of a single report at the destination. Seven cells crossed
   used to produce one motion event. Invisible to an application that only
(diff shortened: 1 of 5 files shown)


vyncint:
## A limit found by CI, not a flake

The first push made `test (ubuntu-latest)` run for sixteen minutes before I cancelled it, while macOS passed. The cause is a real ceiling and it belonged in the test's claim rather than in its timeouts.

**Linux's `N_TTY_BUF_SIZE` is 4 KB.** A thousand DSR replies are ~6 KB, so they cannot all be delivered before the application reads — and the test script reads only *after* its query loop, so several waits burned their full 30 s deadline in turn. macOS has a larger input queue, which is why it passed there.

So the 1000 case was asserting something no terminal can do. 400 replies are ~2.4 KB and fit on both platforms, and 200/400 are exactly the sizes the issue measured losses at (94 and 161 answered), so the test still pins the bug precisely.

The real ceiling is now stated in the test and the changelog: **past roughly 680 replies the application has to read as it asks** — which is true of a real terminal too, and is a property of the tty, not of termlens.

One other fix in the same push: the non-reading child is now `exec sleep 30` rather than `while true; do sleep 5; done`. The loop left a **grandchild** holding the terminal after `Drop` killed the shell; `exec` makes the sleeping process the child itself, so teardown is clean.

PR #155, opened 2026-08-23
fix: support tests without default features

### LINKED ISSUE #143: The crate does not compile with `--no-default-features`, and no CI job would have noticed
**Today** — the crate does not compile with its default features off. Measured against `main` at 0.6.0:

```
$ cargo test --workspace --no-default-features
error[E0433]: cannot find `assert_screen_snapshot` in `termlens`
  --> crates/termlens/src/lib.rs:88:11
   |
88 | termlens::assert_screen_snapshot!(t.screen()); // …or the bundled macro
   |           ^^^^^^^^^^^^^^^^^^^^^^ could not find `assert_screen_snapshot` in `termlens`
   |
note: found an item that was configured out
  --> crates/termlens/src/lib.rs:136:14
```

The crate-level doctest uses the `insta`-gated macro without being gated itself. `test result: FAILED. 20 passed; 1 failed.`

**Why it is worth fixing** — `default = ["insta"]` is documented as optional: *"Disable if you bring your own snapshot tool."* Anyone who takes that offer gets a crate whose own doctests do not build, which reads as "this configuration is unsupported" — and functionally it is, because nothing has ever built it.

The doctest is the symptom. **The cause is that no CI job builds this configuration**: `clippy` and `test` both pass `--all-features`, `check` uses the defaults, and there is no `--no-default-features` anywhere. A feature the README offers is verified by nothing, so this was always going to be found by a user rather than by us.

**Fix** — gate the doctest line on the feature (`#[cfg(feature = "insta")]` inside the example, or split the example so the always-available `insta::assert_snapshot!` form is the ungated one), and add a `--no-default-features` build to `ci.yml` so the configuration cannot rot again. The second half is the part that matters; without it the next feature-gated example reintroduces this.

**Done when** — `cargo test --workspace --no-default-features` passes, a CI job runs it, and `required-green` waits on that job.

pttydou:
## What & why

Fixes #143.

The crate-level snapshot example compiled `termlens::assert_screen_snapshot!`
when the `insta` feature was disabled, even though that macro is feature-gated.
The example now compiles the bundled macro call only with `insta`.

CI also runs `cargo test --workspace --no-default-features` in its own required
job so the supported configuration stays covered. The direct
`insta::assert_snapshot!` example and default/all-features behavior are
unchanged.

## Validation

- `cargo check --workspace --all-targets --no-default-features --target x86_64-unknown-linux-gnu`
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features --target x86_64-unknown-linux-gnu -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features --target x86_64-unknown-linux-gnu`
- `cargo +1.85.0 check --workspace --locked`
- `zizmor --persona=pedantic .github/workflows/`
- `cargo deny check`

The target crate-level doctest compiles with both `--no-default-features` and
`--all-features`. I could not complete the runtime doctest/workspace suites on
Windows because existing tests and examples require Unix APIs and `sh`; the new
CI job runs the full no-default-features command on Ubuntu.

## Checklist

- [x] Snapshot changes (if any) were reviewed with `cargo insta review`, not blind-accepted — none

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,6 +61,19 @@ jobs:
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
       - run: cargo test --workspace --all-features
 
+  no-default-features:
+    name: no-default-features
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # v1
+        with:
+          toolchain: stable
+      - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
+      - run: cargo test --workspace --no-default-features
+
   msrv:
     name: msrv
     runs-on: ubuntu-latest
@@ -136,7 +149,7 @@ jobs:
   required-green:
     name: required-green
     if: always()
-    needs: [fmt, clippy, test, msrv, docs, deny, zizmor]
+    needs: [fmt, clippy, test, no-default-features, msrv, docs, deny, zizmor]
     runs-on: ubuntu-latest
     steps:
       - name: Verify every needed job succeeded
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ listed under a **Changed** or **Removed** heading.
 
 ## [Unreleased]
 
+### Fixed
+
+- **The crate's doctests build with default features disabled.** The bundled
+  snapshot macro example is compiled only when its `insta` feature exists,
+  and CI now runs `cargo test --workspace --no-default-features` so this
+  supported configuration cannot silently rot again.
+
 ## [0.6.1] - 2026-08-23
 
 ### Fixed
--- a/crates/termlens/src/lib.rs
+++ b/crates/termlens/src/lib.rs
@@ -84,6 +84,7 @@
 //! # fn main() -> termlens::Result<()> {
 //! # let mut t = termlens::Terminal::builder().spawn("true")?;
 //! insta::assert_snapshot!(t.screen());        // plain insta…
+//! # #[cfg(feature = "insta")]
 //! termlens::assert_screen_snapshot!(t.screen()); // …or the bundled macro
 //! # Ok(())
 //! # }


vyncint (approved):
This is a good fix, and it does the half of #143 that actually mattered.

## What I checked

The interesting question with `#[cfg(feature = "insta")]` inside a doctest is whether it resolves against **termlens's** features or against the doctest's own crate — because if it were the latter, the line would be stripped *always*, the example would compile under `--no-default-features`, and it would silently stop testing the bundled macro. That failure would look exactly like success.

So I broke the macro call on purpose and ran both configurations:

```
with `termlens::assert_screen_snapshot!(THIS_DOES_NOT_EXIST)`:
  --all-features        error[E0425]: cannot find value `THIS_DOES_NOT_EXIST`
  --no-default-features test result: ok. 21 passed
```

Exactly right. The macro is still covered where the feature is on, and genuinely absent where it is off. Hiding the attribute behind `# ` so the rendered example is unchanged for a reader is a nice touch.

I also ran the new job's command over the whole workspace — `cargo test --workspace --no-default-features`, 176 tests, all green — because it becomes a **required** check via `required-green`, and a new required job that fails on something unrelated blocks every future PR. It does not.

`fmt` and `clippy --all-features` are clean here too; the unticked box in your checklist looks like a missed tick rather than a missed step.

## On the CI job specifically

Wiring it into `required-green` rather than leaving it standalone is the right call — that is the context the branch ruleset actually requires, so the new job is enforced rather than merely present. This is the part #143 said mattered most, and it is the part that stops the next feature-gated example reintroducing the same thing.

## One thing for later, not for this PR

The same gap exists one door down: the `docs` job runs `--all-features`, so `cargo doc` under *default* features has two unresolved intra-doc links to `GraphicsPayload::decode` that nobody sees. Pre-existing on `main`, unrelated to your change, and I have confirmed it is not something you introduced — but it is the same shape of problem you just fixed. I will file it; say the word if you would like it.

Thank you for this.

vyncint:
Merged — thank you, @pttydou. This is your first contribution here and it is a properly good one.

Two things I want to call out, because they are the difference between a fix and a fix that lasts:

**You fixed the cause, not just the symptom.** #143 could have been closed by gating one line, and the doctest would have gone green. It would also have been back within a release, on the next feature-gated example somebody wrote. Adding `--no-default-features` to CI *and* wiring it into `required-green` is what stops that — `required-green` is the context the branch ruleset actually requires, so the job is enforced rather than merely present.

**Your validation list was accurate.** I re-ran all of it and it held up, including the parts that are easy to be optimistic about. The `--no-default-features` suite is 176 tests across the workspace and every one passes, which matters because a new *required* job that fails on something unrelated blocks every future PR.

Your note about not being able to run the Windows suite was the right call too — those tests need Unix APIs and `sh`, so there was nothing to run, and saying so beats a checkbox ticked on faith.

The change is on `main` and will go out in the next release.

I filed #156 for a gap in the same shape that your PR made visible: the `docs` job also runs `--all-features`, so `cargo doc` under *default* features has unresolved intra-doc links that nobody sees. It predates your change entirely — I checked `main` without your commit to be sure — but I would not have gone looking if you had not just shown that "no CI job builds this configuration" is a class of bug here rather than a one-off. It is yours if you want it; equally happy for you to leave it.

PR #167, opened 2026-08-24
feat: add TerminalBuilder::envs

### LINKED ISSUE #164: `TerminalBuilder` has `args` but no `envs`
**Good first issue.** A small, purely additive API method with an existing model
to copy line for line.

### Today

`TerminalBuilder` mirrors `std::process::Command` for arguments — singular and
plural:

```rust
pub fn arg(mut self, arg: impl AsRef<OsStr>) -> Self          // terminal.rs:983
pub fn args<I, S>(mut self, args: I) -> Self                  // terminal.rs:990
```

For the environment it has the singular and the clear, but no plural:

```rust
pub fn env(mut self, key: impl AsRef<OsStr>, value: impl AsRef<OsStr>) -> Self  // :1005
pub fn env_clear(mut self) -> Self                                             // :1019
```

`Command` has `envs`. This does not. So a test setting several variables writes:

```rust
Terminal::builder()
    .env("NO_COLOR", "1")
    .env("PAGER", "cat")
    .env("LANG", "C")
```

and a test that has its variables in a map or a `Vec` has to loop by hand,
rebinding the builder each time, which is exactly the awkwardness `args` was
given a plural form to avoid.

### Why it is small

The field it appends to is already a `Vec<(OsString, OsString)>` named `envs`.
The method is `args` with a pair instead of a scalar:

```rust
/// Set several environment variables for the child.
#[must_use]
pub fn envs<I, K, V>(mut self, vars: I) -> Self
where
    I: IntoIterator<Item = (K, V)>,
    K: AsRef<OsStr>,
    V: AsRef<OsStr>,
{
    self.envs.extend(
        vars.into_iter()
            .map(|(k, v)| (k.as_ref().to_os_string(), v.as_ref().to_os_string())),
    );
    self
}
```

That signature accepts `[( "A", "1" ), ("B", "2")]`, a `HashMap`, a `Vec` of
pairs, and `std::env::vars()`. Worth checking each of those actually compiles —
that is the interesting part of the change, not the body.

### Please also cover

- **Order and duplicates.** `env` appends, so a later call wins at spawn.
  `envs` must behave the same, and mixing `env` and `envs` must respect call
  order. Assert it rather than assuming it.
- **`env_clear` interaction.** Note the doc on `env_clear` (terminal.rs:1019):
  unlike `Command::env_clear`, entries set here *survive* it. Whatever `envs`
  adds has to survive it too, and a test should say so.
- **Docs.** One doc comment in the house style, and a line under `[Unreleased]`
  in `CHANGELOG.md`.

### Testing

`crates/termlens/tests/` drives real programs in real PTYs. A shell that prints
the variables back is enough:

```rust
let mut t = Terminal::builder()
    .envs([("FOO", "1"), ("BAR", "2")])
    .args(["-c", r#"echo "$FOO-$BAR"; read done"#])
    .spawn("sh")?;
t.wait_until(|s| s.contains("1-2"))?;
```

### A note on versioning

This is additive, so it is a **minor** release for a `0.x` crate: it lands on
`main` and ships in 0.7, not in a 0.6 patch. `cargo-semver-checks` runs at release
time and will agree. Nothing for you to do beyond knowing why the milestone says
what it says.

### Done when

`TerminalBuilder::envs` exists, is documented, is tested for ordering and for the
`env_clear` interaction, and `CHANGELOG.md` mentions it.

### Getting started

`CONTRIBUTING.md` has the setup and the DCO sign-off (`git commit -s`); AI help
is welcome, AI attribution is not. Questions here are welcome at any stage.

k4its1t:
## What & why

Add `TerminalBuilder::envs` as the plural counterpart to `env`, accepting any iterator of key-value pairs while preserving iteration and builder-call order. This removes the need to manually rebind a builder when variables already live in an array, `Vec`, `HashMap`, or environment iterator.

The integration coverage verifies duplicate-key precedence across mixed `env`/`envs` calls and confirms that values added both before and after `env_clear` survive while inherited variables remain blocked.

Closes #164

## Verification

- `cargo test -p termlens --test basic` (13 passed)
- `cargo test --workspace --all-features`
- `cargo test --workspace --no-default-features`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps`
- `RUSTDOCFLAGS="-D warnings" cargo doc --no-deps --all-features`
- `cargo +1.85.0 check --workspace --locked`
- `.github/scripts/check-dco.sh upstream/main..HEAD`
- `.github/scripts/check-no-ai-attribution.sh upstream/main..HEAD`

The stress workflow was not run because this change does not touch wait or timing behavior. No snapshots changed.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ listed under a **Changed** or **Removed** heading.
 
 ## [Unreleased]
 
+### Added
+
+- **`TerminalBuilder::envs` sets several child environment variables from an
+  iterator of key-value pairs.** Values keep their iteration and builder-call
+  order, and remain explicit when `env_clear` disables inherited variables.
+
 ### Fixed
 
 - **The crate's doctests build with default features disabled.** The bundled
--- a/crates/termlens/src/terminal.rs
+++ b/crates/termlens/src/terminal.rs
@@ -1008,13 +1008,35 @@ impl TerminalBuilder {
         self
     }
 
+    /// Set several environment variables for the child.
+    ///
+    /// Variables are applied in iteration order, so a later value for the
+    /// same key wins, including across calls to [`env`](Self::env). Entries
+    /// set here always reach the child, regardless of call order relative to
+    /// [`env_clear`](Self::env_clear).
+    #[must_use]
+    pub fn envs<I, K, V>(mut self, vars: I) -> Self
+    where
+        I: IntoIterator<Item = (K, V)>,
+        K: AsRef<OsStr>,
+        V: AsRef<OsStr>,
+    {
+        self.envs.extend(
+            vars.into_iter()
+                .map(|(key, value)| (key.as_ref().to_os_string(), value.as_ref().to_os_string())),
+        );
+        self
+    }
+
     /// Don't inherit the parent process environment: the child sees only
-    /// variables set via [`env`](Self::env) (plus the default `TERM`, see
-    /// [`spawn`](Self::spawn)). Strict control keeps tests hermetic — a
-    /// developer's exotic `LS_COLORS` should never change a snapshot.
+    /// variables set via [`env`](Self::env) or [`envs`](Self::envs) (plus the
+    /// default `TERM`, see [`spawn`](Self::spawn)). Strict control keeps tests
+    /// hermetic — a developer's exotic `LS_COLORS` should never change a
+    /// snapshot.
     ///
     /// Note this differs from `std::process::Command::env_clear`, which also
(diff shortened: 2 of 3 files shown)


vyncint (approved):

vyncint:
Thanks for the contribution, @k4its1t! 

The implementation looks very clean and well-structured, and I really appreciate the thorough integration test coverage and checklist diligence. Merging this now! 🚀

PR #196, opened 2026-08-28
fix: normalize colon-form SGR colours

### LINKED ISSUE #180: A colon-form SGR colour (`38:2::r:g:b`) is silently dropped
**Today** — the colon-separated form of an RGB colour is dropped. Reproduced
against **0.7.0**, both spellings asking for the same colour:

```sh
printf '\033[38;2;10;20;30mA\033[0m'      # semicolon form
printf '\033[38:2::10:20:30mB\033[0m'     # colon form
```

```
cell(0,0).style().fg == Rgb(10, 20, 30)   # A  — correct
cell(0,1).style().fg == Default           # B  — the colour is gone
```

`4:3` (curly underline) is dropped the same way: `underline == false`.

### Why it is worth fixing

The colon form is not exotic. It is the ISO 8613-6 / ITU T.416 spelling that
`38;2;…` was always a simplification of, and it is what a number of
applications and libraries emit — it is unambiguous where the semicolon form
is not, which is the whole reason it exists.

The failure mode is the one this crate exists to prevent, and says so. From
`emu/shadow.rs`, on the three attributes vt100 drops:

> three distinct renderings collapse into one value, and a test asserting that
> a password field is masked **passes against an application that prints the
> secret in clear**.

A colour that silently becomes `Default` is the same shape: a test asserting
`fg == Rgb(…)` fails against an application that is right, and a test asserting
`fg == Default` **passes against an application that set a colour**. Both are
wrong answers presented as confident ones.

The crate already knows this spelling exists. `rewrite_sgr` in `emu/shadow.rs`
handles it explicitly, with a comment — *"Colon form (`38:2::r:g:b`) is one
self-contained group"* — because it has to step over the sub-parameters to
avoid mistaking the `2` for a blink. So the shadow parser understands the form
and the primary one does not.

### Fix

`vt100` is the primary parser and does not decode sub-parameters into a colour.
Options, in increasing order of ambition:

1. **Document it** in `README.md`'s "Known limitations", which is careful about
   this class of gap and currently says nothing about SGR sub-parameters. Worth
   doing on its own.
2. **Normalise before the grid sees it** — rewrite a complete
   `38:2:…`/`48:2:…`/`38:5:…` group into its semicolon equivalent in the byte
   stream. `emu/shadow.rs` is the precedent for that shape and its module doc
   argues why rewriting complete SGR sequences is sound; the same argument
   applies, and the same care is needed.

Note the empty field: `38:2::r:g:b` has a colour-space slot between the `2` and
the `r` which is usually empty, and `38:2:r:g:b` (without it) also occurs in
the wild. Handle both or say which you handled.

### Tests

`tests/styles.rs` drives styles through a real PTY and is where a case belongs;
`src/emu/shadow.rs`'s tests already cover the colon form on the rewriter side
and show the shape.

### Done when

Either the limitation is documented, or both spellings of the same colour
produce the same `Style`, with a test that fails if one of them regresses.

### Getting started

`CONTRIBUTING.md` has the setup and the DCO sign-off (`git commit -s`); AI help
is welcome, AI attribution is not. The documentation-only version is a fine
first PR.

anisayakmitra-in:
## Summary

- normalize supported colon-form SGR colours before primary-grid parsing
- retain non-empty colour-space forms and unrecognised control sequences unchanged
- add direct byte-level coverage for both conversion and preservation behaviour

Closes #180.

## Testing

- cargo test -p termlens --lib
- .github/scripts/check-dco.sh main..HEAD
- .github/scripts/check-no-ai-attribution.sh main..HEAD

The full package test target includes Unix-only process integration tests and does not compile on Windows; the portable library suite passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ listed under a **Changed** or **Removed** heading.
 
 ### Fixed
 
+- **Colon-form SGR colours now reach cell styles.** `38:2::r:g:b`,
+  `38:2:r:g:b`, and indexed foreground and background colours are normalized
+  before the backend parses them, matching their semicolon-form equivalents.
+
 - **Mouse `click` / `click_with` / `drag` / `scroll` refuse coordinates
   outside the current grid.** A real terminal cannot produce an off-window
   mouse event; sending one was the same class of mistake as clicking with
--- a/crates/termlens/src/emu/shadow.rs
+++ b/crates/termlens/src/emu/shadow.rs
@@ -338,6 +338,16 @@ mod tests {
         String::from_utf8_lossy(&out).replace('\x1b', "E")
     }
 
+    /// The primary rewriter's output, as a readable string.
+    fn normalized(bytes: &[u8]) -> String {
+        let mut state = State::Ground;
+        let mut pending = Vec::new();
+        let mut out = rewrite_stream(&mut state, &mut pending, bytes, normalize_colon_colours);
+        // Anything still held at the end of the stream is not a complete SGR.
+        out.append(&mut pending);
+        String::from_utf8_lossy(&out).replace('\x1b', "E")
+    }
+
     #[test]
     fn the_three_dropped_attributes_get_carriers() {
         assert_eq!(shadowed(b"\x1b[5mX"), "E[1mX"); // blink -> bold
@@ -404,6 +414,28 @@ mod tests {
         ] {
             assert_eq!(
                 shadowed(seq),
+                String::from_utf8_lossy(seq).replace('\x1b', "E"),
+                "sequence {:?} must pass through untouched",
+                String::from_utf8_lossy(seq)
+            );
+        }
+    }
+
+    #[test]
+    fn colon_colour_normalization_preserves_unrecognized_bytes() {
+        assert_eq!(normalized(b"\x1b[38:2::10:20:30m"), "E[38;2;10;20;30m");
+        assert_eq!(normalized(b"\x1b[38:5:196m"), "E[38;5;196m");
+        assert_eq!(normalized(b"\x1b[1;38:2::4:5:6;3m"), "E[1;38;2;4;5;6;3m");
+
+        for seq in [
+            &b"\x1b[4:3m"[..],
+            &b"\x1b[38:2::9m"[..],
+            &b"\x1b[38:2:1:10:20:30m"[..],
+            &b"\x1b[?25l"[..],
+            &b"\x1b]8;;http://x/\x1b\\"[..],
+        ] {
+            assert_eq!(
+                normalized(seq),
                 String::from_utf8_lossy(seq).replace('\x1b', "E"),
                 "sequence {:?} must pass through untouched",
                 String::from_utf8_lossy(seq)
--- a/crates/termlens/src/emu/shadow.rs
+++ b/crates/termlens/src/emu/shadow.rs
@@ -57,6 +57,32 @@ enum State {
     Csi,
 }
 
+/// Rewrites colon-form extended colours into the semicolon form understood by
+/// `vt100`. Like [`AttrShadow`], it only holds a complete plain SGR and emits
+/// every other sequence unchanged.
+pub(super) struct ColorNormalizer {
+    state: State,
+    pending: Vec<u8>,
+}
+
+impl ColorNormalizer {
+    pub(super) fn new() -> Self {
+        Self {
+            state: State::Ground,
+            pending: Vec::new(),
+        }
+    }
+
+    pub(super) fn feed(&mut self, bytes: &[u8]) -> Vec<u8> {
+        rewrite_stream(
+            &mut self.state,
+            &mut self.pending,
+            bytes,
+            normalize_colon_colours,
+        )
+    }
+}
+
 /// A parallel `vt100::Parser` carrying blink, conceal and strikethrough.
 pub(super) struct AttrShadow {
     parser: ::vt100::Parser,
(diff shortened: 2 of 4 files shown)


vyncint (changes requested):
Thank you for this — the approach is right, and the part I was most worried
about when I wrote #180 is the part you got right.

**The implementation is correct.** I did not take that on trust; I drove ten
shapes through a real PTY on your branch:

| input | result |
|---|---|
| `38;2;10;20;30` (baseline) | `Rgb(10,20,30)` |
| `38:2::10:20:30` | `Rgb(10,20,30)` |
| `38:2:10:20:30` | `Rgb(10,20,30)` |
| `48:2::1:2:3` | `bg = Rgb(1,2,3)` |
| `38:5:196` / `48:5:21` | `Indexed(196)` / `Indexed(21)` |
| `1;38:2::4:5:6;3` | `Rgb(4,5,6)` **and bold and italic both survive** |
| `38:2:1:10:20:30` (non-empty colour space) | left alone, not corrupted |
| `4:3` (curly underline) | unchanged |
| `38:2::9` (malformed) | unchanged |

The mixed-parameter row is the one that matters most: your rewriter takes a
sequence apart into groups and rebuilds it, so the surrounding parameters are
exactly what a careless version would drop. They survive, in order.

Three other things worth saying:

- **Extracting `rewrite_stream` and parameterising it by the rewrite function
  is the right call.** The state machine is the subtle part — held bytes, ESC
  abandoning a partial sequence, control bytes inside a CSI — and reusing the
  proven one instead of writing a second is what makes this safe to put in
  front of the *primary* parser.
- **Routing raw bytes to the shadow and normalized bytes to the grid keeps the
  correspondence invariant.** `snapshot`'s `debug_assert_eq!` on the two grids
  ran on every snapshot across 336 tests in debug and stayed quiet.
- **It costs nothing.** I measured 40k lines through an 80x24 grid, best of 5,
  release: **130.5 ms on your branch vs 130.4 ms on `main`** plain, 87.4 vs 86.6
  with four SGRs a line. The third pass is below the noise floor, which matches
  what `shadow.rs` says about the second one.

---

### One change before merge

`normalize_colon_colours` has no direct unit test. Everything above is
exercised only through two `vt100.rs` tests and one integration test, all of
which assert the *happy path* — a colour arrives.

The property that makes this approach safe is the other one: **everything that
is not a recognised colour group passes through byte-for-byte.** That is what
lets a rewriter sit in front of the real grid rather than a shadow, and right
now nothing would fail if it broke. A future edit to the `match` could start
eating a parameter and every existing test would stay green.

`shadow.rs` already has the pattern — the `shadowed()` helper at the bottom of
the test module returns the rewritten bytes as a readable string, which makes
byte-level assertions one line each. A `normalized()` twin would give you:

```rust
// recognised forms are rewritten
assert_eq!(normalized(b"\x1b[38:2::10:20:30m"), "E[38;2;10;20;30m");
assert_eq!(normalized(b"\x1b[38:5:196m"),       "E[38;5;196m");
// surrounding parameters survive, in order
assert_eq!(normalized(b"\x1b[1;38:2::4:5:6;3m"), "E[1;38;2;4;5;6;3m");
// everything else is untouched
assert_eq!(normalized(b"\x1b[4:3m"),             "E[4:3m");
assert_eq!(normalized(b"\x1b[38:2::9m"),         "E[38:2::9m");
assert_eq!(normalized(b"\x1b[38:2:1:10:20:30m"), "E[38:2:1:10:20:30m");
assert_eq!(normalized(b"\x1b[?25l"),             "E[?25l");
assert_eq!(normalized(b"\x1b]8;;http://x/\x1b\\"), "E]8;;http://x/E\\");
```

That last one is worth having explicitly: an `OSC 8` URI is full of colons and
now flows through this code, so a test saying it comes out unchanged is cheap
insurance for a feature that would otherwise break quietly.

Please break one of them once and check it goes red before you push — the house
rule is that a test has to be able to fail, and for a preservation guarantee
that is easy to get wrong.

### Two notes, neither blocking

- The PR description has literal `\n` and `\` escapes in it, so it renders as
  one run-on line. The commit messages themselves are clean, so nothing reaches
  the history — worth fixing only so the PR page reads properly.
- Not handling a non-empty colour space (`38:2:1:…`) is the right call, and you
  said which forms you handled, which is what #180 asked for. No change wanted;
  just confirming it was noticed and is deliberate.

Everything else here is ready. Thanks again — this is a careful piece of work on
a genuinely fiddly part of the crate.

anisayakmitra-in:
Added direct byte-level tests for the colon SGR normalizer, covering recognised RGB and indexed forms, mixed surrounding parameters, and byte-for-byte preservation of malformed, non-empty-colour-space, private CSI, and OSC 8 sequences. The new test was verified to fail with normalization disabled, then pass after restoration. Validation: cargo test -p termlens --lib (150 tests), DCO, and attribution checks. The PR description has also been re-rendered without literal escape sequences.

vyncint (approved):
This is exactly right, and the test is better than what I suggested.

Writing the preservation loop as "the output equals the input, rendered the
same way" instead of hand-written expected strings means the assertions cannot
drift from what they claim, and the failure names the offending sequence. I
asked for eight literals; you wrote something that reads better and covers more.

I checked your "verified to fail with normalization disabled" claim, and then
checked the half it does not reach:

```
A: normalization disabled
   caught — left: "E[38:2::10:20:30m"  right: "E[38;2;10;20;30m"

B: rewriter made over-eager (any colon group treated as a colour)
   caught — sequence "\u{1b}[4:3m" must pass through untouched
            left: "E[4;3m"  right: "E[4:3m"
```

Worth pointing out because it is the interesting part: **disabling
normalization leaves every preservation assertion passing trivially** —
unchanged is unchanged. Only an over-eager rewriter exercises them, and that is
the direction that would corrupt a real screen. Your test catches both. That is
the property I wanted pinned, and it is now pinned from both sides.

The rest re-verified on the updated branch: 337 tests green, CI green on both
OSes, and commit `8e522b8` touches only `mod tests` — so the production code is
byte-identical to the revision I put through ten PTY shapes last round,
including `1;38:2::4:5:6;3` keeping its bold and italic. Nothing to re-test
there.

Thanks for the quick turnaround, for the re-rendered description, and for
breaking it once before pushing rather than telling me it would fail. Merging.

Issue #149, opened 2026-08-22
Unix only: what Windows support would actually take


vyncint:
**Today** — Unix only. Linux and macOS in CI; a Windows user cannot run the crate at all.

**Why it is worth fixing** — it is the one asterisk on the crate's premise. termlens tests *terminal programs*, and a great many of them are cross-platform and broken on Windows in ways their authors cannot see, which is exactly the audience a PTY harness is for. It is also the largest single thing standing between the current API and a 1.0 that means what it says.

**Status of the ground** — `portable-pty` already supports ConPTY, so the PTY layer is not the obstacle. What needs deciding, roughly in order of difficulty:

- **`Terminal::signal`** is `#[cfg(unix)]` and models `kill(2)`. Windows has no equivalent; the honest answer is probably that the method stays Unix-only rather than being emulated into something that does not mean the same thing.
- **The PTY lifecycle lock** exists for a macOS `revoke()` behaviour that has no Windows analogue. It should not become a Windows tax by accident, and its motivation is now recorded in two places (`PTY_LIFECYCLE`, and the retry added in #140).
- **Fixtures** are `/bin/sh` scripts throughout. The suite needs a portable way to say "print this, then wait" before any of it can run.
- **What ConPTY does to the byte stream.** It is not a passthrough: it renders and re-emits, so the escape sequences arriving at the emulator are not necessarily the ones the application wrote. That is the part most likely to make a feature — synchronized-update frames above all, and the query responder — mean something different on Windows, and it wants investigating before anything is promised.

**Worth noting for the record** — the v0.4 and v0.5 milestones both list "Unix only" under *out of scope as inherent*, while the README says Windows is "planned, not designed out". Those two statements do not agree, and the disagreement is worth settling on purpose. This issue is filed unmilestoned for that reason: it is tracked, not scheduled.

**Done when** — either the suite runs on Windows in CI, or the README stops saying it is planned.

vyncint:
Status at v0.9.0: **settled on purpose, not started in code** — in milestone v0.10 as four ordered steps, each shippable alone.

1. **Rust fixtures replace `/bin/sh`** — #249. Worth doing for Unix alone (the suite spawns a shell at 98 sites), and the precondition for running the suite anywhere without one.
[...]

PR #197, opened 2026-08-28
ci: cover decode-only configuration

### LINKED ISSUE #161: CI never builds `--no-default-features --features decode`, and clippy only runs `--all-features`
**Good first issue.** No prior knowledge of the codebase needed — this one is
entirely in `.github/workflows/ci.yml`, and the fix is currently *green*, so you
cannot break anything by adding it.

### Today

The crate has two optional features, `insta` (on by default) and `decode` (off).
That is four possible builds. CI builds three of them:

| configuration | built by |
|---|---|
| `--all-features` | `clippy`, `test`, `docs` |
| default (`insta` only) | `docs` (`cargo doc --no-deps`, ci.yml:109) |
| `--no-default-features` | `no-default-features` (`cargo test`, ci.yml:75) |
| **`--no-default-features --features decode`** | **nothing** |

And `clippy` only ever runs one of them — `--all-features`, ci.yml:43 — so a lint
that fires in any other configuration goes unnoticed too.

### Why it is worth fixing

This is the third time the same shape has bitten. #143: the crate did not compile
with `--no-default-features`, and no job would have noticed. #156: `cargo doc`
warned under *default* features, and no job would have noticed. Both were found by
a person running a configuration CI did not.

`decode` without `insta` is a real configuration — it is what a project that
brings its own snapshot tool and wants to decode kitty payloads asks for. Nothing
verifies it compiles.

### It is green right now

Measured on `main` today, so this is guard work, not repair work:

```sh
cargo clippy --workspace --all-targets --no-default-features -- -D warnings
cargo clippy --workspace --all-targets --no-default-features --features decode -- -D warnings
cargo test --workspace --no-default-features --features decode --no-run
```

All three pass. The point of the change is that they keep passing.

### Fix

In `.github/workflows/ci.yml`:

1. Have the `no-default-features` job also build the `decode`-only combination —
   `cargo test --workspace --no-default-features --features decode` — or add a
   sibling job for it, whichever reads better to you.
2. Run clippy over the configurations too, not only `--all-features`.

If you add a **new job**, remember `required-green` (ci.yml:150) lists the jobs it
waits for. A job missing from that list is a job whose failure merges anyway,
which is the exact hole #143 left. Add it there, and if it should gate `main`, say
so in the PR — the branch ruleset needs the context name and only a maintainer can
add it.

Keep an eye on job count: each one is a fresh runner. Folding the extra
configurations into the existing jobs is a perfectly good answer, and probably
the better one.

### Done when

`--no-default-features --features decode` is built by CI, clippy covers more than
`--all-features`, and anything new is wired into `required-green`.

### Getting started

`CONTRIBUTING.md` has the setup, the DCO sign-off (`git commit -s`) and the AI
policy — **AI help is welcome, AI attribution is not**. Ask anything here; a
half-finished PR with a question in it is very welcome.

anisayakmitra-in:
Fixes #161. Extends the existing no-default-features job to test the decode-only build, and extends the existing clippy job to lint both no-default configurations. Keeping the checks in their current jobs preserves the required-green wiring and avoids extra runners. Verification: cargo test -p termlens --lib --no-default-features --features decode; cargo clippy -p termlens --lib --no-default-features -- -D warnings; cargo clippy -p termlens --lib --no-default-features --features decode -- -D warnings; git diff --check; DCO and attribution checks. I did not run the full all-targets command locally because this Windows host cannot compile the repository's Unix-only integration test target; the PR CI runs it on Linux.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,8 @@ jobs:
           components: clippy
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
       - run: cargo clippy --workspace --all-targets --all-features -- -D warnings
+      - run: cargo clippy --workspace --all-targets --no-default-features -- -D warnings
+      - run: cargo clippy --workspace --all-targets --no-default-features --features decode -- -D warnings
 
   # CI runners have no TTY; termlens creates its own PTYs. This suite passing
   # headless is the proof of the crate's core value claim.
@@ -73,6 +75,7 @@ jobs:
           toolchain: stable
       - uses: Swatinem/rust-cache@6323deb102c322ba6fcbdcafc7e3dddab59af2b6 # v2
       - run: cargo test --workspace --no-default-features
+      - run: cargo test --workspace --no-default-features --features decode
 
   msrv:
     name: msrv


vyncint (approved):
Thank you — this is exactly the shape the issue was hoping for, and it closes a
gap that had bitten three times.

You took the option the issue suspected was better and I agree: folding the
configurations into the existing `clippy` and `no-default-features` jobs rather
than adding runners. That also means `required-green` needs no change, which is
the trap #143 left and the one I was most curious whether this would step in.
It didn't.

I checked that the new commands actually *ran* rather than trusting the green
tick, since a CI change that silently doesn't execute is the failure mode here:

```
clippy job:
  cargo clippy --workspace --all-targets --all-features -- -D warnings
  cargo clippy --workspace --all-targets --no-default-features -- -D warnings
  cargo clippy --workspace --all-targets --no-default-features --features decode -- -D warnings

no-default-features job:
  cargo test --workspace --no-default-features
  cargo test --workspace --no-default-features --features decode
```

All five ran, all passed. All four feature combinations are now built, and
clippy covers three of them instead of one.

Also appreciated: you said plainly that you couldn't run the full `--all-targets`
command locally because the host is Windows and the integration tests are
Unix-only. That's the right call and the right thing to write down — it's
accurate (the crate is Unix-only for now, see #149), and CI covered it on Linux
and macOS.

Merging. Thanks again.

PR #207, opened 2026-08-29
fix(ci): run install check after releases

### LINKED ISSUE #206: install.yml never runs after a release: the release event comes from GITHUB_TOKEN
## Summary

`install.yml` is meant to run automatically after every release:

```yaml
on:
  release:
    types: [published]
```

It never has. The release is created by `release.yml`'s `github-release` job
with `gh release create … --verify-tag` under the workflow's own
`GITHUB_TOKEN`, and GitHub does not fire event triggers for actions taken
with that token — the standard guard against recursive workflows. So the
`release: published` event that `install.yml` listens for is produced by
exactly the actor whose events cannot trigger it.

## Evidence

The workflow's run history has only `workflow_dispatch` and `schedule` runs:

| release | `install.yml` run for it |
|---|---|
| v0.7.0 (2026-08-27 01:5x) | `workflow_dispatch` at 02:06, started by hand |
| v0.8.0 (2026-08-29 01:43) | none until a `workflow_dispatch` was started by hand |

Both times the consumer check ran only because the person releasing
remembered to start it. The weekly schedule would have caught a broken publish
eventually — up to six days late.

## Why it matters

`install.yml` is the only gate that asks what crates.io actually serves: a
file missing from the archive, a fresh dependency resolution, a feature that
resolves in this workspace and nowhere else. Its own header says those three
failures live "in that gap and nowhere else". A gate that has to be started by
hand after every release is a gate that will be skipped on the release where
it mattered.

## Fix

Have `release.yml` start it explicitly instead of relying on the event. Two
options:

1. **`workflow_call`.** Add `on: workflow_call: inputs: version` to
   `install.yml` and a job in `release.yml` after `github-release`:
   ```yaml
   install:
     needs: github-release
     uses: ./.github/workflows/install.yml
     with:
       version: ${{ github.ref_name }}
   ```
   Runs inside the release run, shows up in its job list, needs no extra
   permissions. The `version` input already exists for `workflow_dispatch`, so
   the reusable form is a small change; the `concurrency` group and the
   `inputs.version` references need checking under `workflow_call`.
2. **Dispatch it.** A step in `github-release` running
   `gh workflow run install.yml -f version="${GITHUB_REF_NAME#v}"`, which
   needs `actions: write` on that job's token.

(1) is the better shape: the consumer check becomes part of the release run
rather than a separate run somebody has to look for. Either way, keep the
`release:` trigger too — it costs nothing and would work for a release created
by hand in the UI.

Note for `zizmor`: a `workflow_call` job inherits the caller's token; the
reusable workflow's own `permissions: contents: read` still applies.

## Done when

Pushing a `v*` tag results in `install.yml` running against the published
version without anyone starting it, and `docs/RELEASING.md` no longer needs
a footnote about it.

---

Found on 2026-08-29 while releasing v0.8.0: waited for the release-triggered
run, saw none, checked the history.

anisayakmitra-in:
## Summary

- make the registry-consumer workflow callable from the release pipeline
- run it after the GitHub Release is created, instead of relying on an event emitted by GITHUB_TOKEN
- normalize a reusable caller's X.Y.Z tag before querying crates.io

Closes #206.

## Verification

- cargo test --workspace --all-features --lib
- cargo fmt --all -- --check
- git diff --check
- .github/scripts/check-dco.sh origin/main..HEAD
- .github/scripts/check-no-ai-attribution.sh origin/main..HEAD

Not run locally: the pinned zizmor audit, because this environment does not provide pipx; CI runs that audit at the repository's configured pedantic level.

### DIFF
--- a/.github/workflows/install.yml
+++ b/.github/workflows/install.yml
@@ -25,6 +25,12 @@ name: install
 on:
   release:
     types: [published]
+  workflow_call:
+    inputs:
+      version:
+        description: "Version to verify (default: the release tag or newest published)"
+        required: false
+        type: string
   workflow_dispatch:
     inputs:
       version:
@@ -58,7 +64,11 @@ jobs:
         run: |
           api="https://crates.io/api/v1/crates/termlens"
           agent="termlens-install-check (github actions)"
-          version="${WANTED:-${TAG#v}}"
+          # A reusable caller has the release tag available, while a manual
+          # dispatch accepts either a tag or the crate version. Normalize both
+          # before querying crates.io.
+          version="${WANTED#v}"
+          version="${version:-${TAG#v}}"
           if [ -z "$version" ]; then
             version=$(curl -sSf -H "User-Agent: $agent" "$api" | jq -r .crate.max_stable_version)
           fi
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -111,6 +111,16 @@ jobs:
             --notes-file "$RUNNER_TEMP/notes.md" \
             --verify-tag
 
+
+  # Releases created with GITHUB_TOKEN do not emit a release event that can
+  # start another workflow. Call the consumer check directly after publishing
+  # the GitHub release so every tag verifies what crates.io serves.
+  install:
+    name: verify published crate
+    needs: github-release
+    uses: ./.github/workflows/install.yml
+    with:
+      version: ${{ github.ref_name }}
   # Tell the repository that tests this tool against a real subject that a
   # new version exists, so the deep run happens now rather than at its next
   # scheduled tick.
--- a/docs/RELEASING.md
+++ b/docs/RELEASING.md
@@ -48,7 +48,9 @@ Pushing the tag runs `release.yml`, which:
 4. `cargo publish -p termlens` via Trusted Publishing (OIDC) — the
    repository stores no tokens,
 5. creates the GitHub Release with notes extracted from the CHANGELOG
-   section for that version (`.github/scripts/extract-changelog.sh`).
+   section for that version (`.github/scripts/extract-changelog.sh`), and
+6. runs the registry-consumer check against that published version on Linux
+   and macOS.
 
 ## If something fails mid-release
 


vyncint (approved):
Thank you — this is a clean fix, and it solves the problem the issue actually described rather than the symptom.

I verified it rather than reading it, since a CI-only change can't be exercised by the suite:

**The version normalization is load-bearing, not cosmetic.** That was worth catching. Running both forms of the resolution over every trigger:

| trigger | `WANTED` | `TAG` | before | after |
|---|---|---|---|---|
| `workflow_call` (this PR) | `v0.8.0` | — | **`v0.8.0`** | `0.8.0` |
| `release: published` | — | `v0.8.0` | `0.8.0` | `0.8.0` |
| `workflow_dispatch` | `0.8.0` | — | `0.8.0` | `0.8.0` |
| `workflow_dispatch` | `v0.8.0` | — | **`v0.8.0`** | `0.8.0` |
| `schedule` | — | — | query `max_stable_version` | same |

Without the `#v` strip, the new caller would have queried `/crates/termlens/v0.8.0`, 404'd, and failed with *"v0.8.0 never appeared on crates.io"* — a confusing way to break the very gate being added. The dispatch path gained tag-form input for free.

**The reusable call is wired correctly.** The concurrency group resolves through `inputs.version` under `workflow_call` (`github.event.release` is null there, since the caller's event is the tag push), giving `install-v0.8.0` — distinct from the caller's `release-refs/tags/v0.8.0`, so there's no self-deadlock. `permissions: contents: read` still restricts the inherited token, no secrets are needed, and `needs: github-release` puts it downstream of `publish`. It also matches the `ci:` reusable call already in that file, which is the right precedent to follow.

**Everything else:** all 11 checks pass including `zizmor --persona=pedantic`; DCO and the attribution policy pass on the commit itself; both workflows parse and the job graph is what it should be; `RELEASING.md` gains step 6 and has no manual footnote left, so the issue's "Done when" is fully met.

One nit, not worth a round trip: `release.yml` has a doubled blank line before the new job and none between its `with:` block and the comment introducing `notify-testing-repo`. The file's convention is comment-above-job so the meaning is unambiguous — I'll tidy the spacing next time that file is touched.

Merging. Thanks for picking up a `good first issue` and doing the careful version of it.

Issue #232, opened 2026-09-03
`DECSC`/`DECRC` do not save or restore the designated character set


vyncint:
**Today** — `DECSC` (`ESC 7`) and `DECRC` (`ESC 8`) do not save or restore which character set is designated. Measured against `main` (5394c11) on a 24x3 terminal, printing `lqk` — the three bytes that draw `┌─┐` when G0 holds DEC Special Graphics:

```
$ printf '\033(0lqk'                       # G0 = line drawing, no reset
row0 = "┌─┐"                               # correct

$ printf '\033(0\0337\033(B\0338lqk'       # designate 0, DECSC, designate B, DECRC
row0 = "lqk"                               # G0 is still ASCII: the restore did nothing
```

`ESC 8` should have put G0 back to the line-drawing set that was current when `ESC 7` ran. The designation is tracked in `crates/termlens/src/emu/seq.rs`, whose module docs already record the boundary: G0/G1 designations and the `SO`/`SI` locking shifts are modelled, and the README's Known limitations says `DECSC`/`DECRC` "do not save or restore the charset state".

**Why it is worth fixing** — save/restore around a cursor move is a common idiom, and full-screen applications reach for it precisely when they are about to draw a border: save, jump, draw the frame, restore. When the restore silently loses the charset, the border after it renders as `lqk` — literal letters where a user sees box drawing. That is the exact failure #204 fixed for the designation itself, arriving through a different door: a test asserting a border is correct fails against an application that is correct, and a snapshot that blessed `lqk` keeps passing after the border breaks.

**Fix** — the tracker already holds the state this needs. Give it a saved slot — the G0 and G1 designations plus which of the two is invoked — and push to it on `ESC 7`, pop from it on `ESC 8`. Two details worth getting right, both cheap:

- `DECSC` with no matching `DECRC`, and `DECRC` with nothing saved. xterm treats a restore with nothing saved as a reset to the default (G0 = ASCII, G0 invoked); doing the same is the least surprising choice and needs no new state.
- A hard reset (`RIS`) must clear the saved slot too, or a `DECRC` after `RIS` resurrects a designation from before the reset. `crates/termlens/tests/charset.rs` already has `a_hard_reset_returns_to_ascii` to extend.

Out of scope here: the cursor position and attributes that `DECSC` also saves. This issue is only the charset half, which is the half the README names.

[...]

PR #239, opened 2026-09-03
fix(emu): model G2/G3 designation and SS2/SS3 single shifts

### LINKED ISSUE #235: G2 and G3 and the single shifts SS2/SS3 are consumed and ignored
**Today** — G2 and G3, and the single shifts that invoke them, are not modelled: the designation is consumed and the shift does nothing. Measured against `main` (5394c11), where `l` is the byte that draws `┌` when a line-drawing set is invoked:

```
$ printf '\033*0\033Nl\033(B|'    row0 = "l|"     # ESC * 0 designates G2, ESC N (SS2) invokes it for one char
$ printf '\033+0\033Ol\033(B|'    row0 = "l|"     # ESC + 0 designates G3, ESC O (SS3) — same
$ printf '\033(0l\033(B|'         row0 = "┌|"     # G0, for contrast: translated
```

`crates/termlens/src/emu/seq.rs`'s module docs record the boundary — "G2/G3 with their single shifts are not modelled" — and so does the README's Known limitations.

**Why it is worth fixing** — a single shift is how an application draws *one* graphic character without disturbing the set it has invoked, which is exactly what a mixed line of text and box-drawing wants. termlens renders the letter instead of the glyph, so the grid shows `l` where the user sees `┌`: the same silent shape as #204, where a test asserting a correct border fails and a snapshot blessing the wrong one keeps passing.

It is also the last unmodelled corner of a mechanism that is otherwise complete, so closing it retires the caveat rather than shrinking it.

**Fix** — the tracker already models two designation slots and which is invoked. This extends both:

- **Designation**: recognise `ESC * Ps` (G2) and `ESC + Ps` (G3) beside the existing `ESC ( Ps` (G0) and `ESC ) Ps` (G1), into two more slots.
- **Single shifts**: `ESC N` (SS2) and `ESC O` (SS3) invoke G2 / G3 for **exactly one** character, then the previous invocation resumes. That one-character lifetime is the part to get right and the part to test hardest — a single shift that stays invoked is worse than one that never was, because it silently translates the rest of the line.

`SO`/`SI` continue to lock-shift between G0 and G1 only; a single shift overrides whichever is locked, for one character. Careful with the shift's interaction with a `RIS` or a `DECSC`/`DECRC` landing mid-shift, and note that `ESC O` is also the prefix of the application-cursor-key encodings (`ESC O A`) — the tracker must not confuse an SS3 in the output stream with a key it would *send*, which is a different direction entirely.

This is the largest of the current good-first-issues: it touches the tracker's state machine rather than a table. `crates/termlens/tests/charset.rs` has four tests to model on, and the pattern to follow is `shift_out_and_shift_in_select_the_designated_set`.

**Done when** — `ESC * 0`, `ESC N`, `l` puts `┌` on the grid and the character *after* it is back to the previously invoked set; the same for G3 via `ESC + 0` and `ESC O`; `SO`/`SI` still behave as they do today; a single shift that is followed by a `RIS` does not survive it; `crates/termlens/tests/charset.rs` covers each; and the README and `seq.rs` module docs no longer list G2/G3 as unmodelled.

cestercian:
## Summary

Fixes #235. `ESC *` / `ESC +` designate G2/G3; `ESC N` (SS2) and `ESC O` (SS3) invoke them for one GL character. The designation was already consumed and the shift did nothing, so a mixed line of text and box-drawing showed `l` where the user sees `┌`.

A pending single shift is consumed on the next GL character (not on intervening C0 / designations), and `RIS` clears it. Locking shifts stay G0/G1 (`SO`/`SI`); `LS2`/`LS3` and 8-bit C1 SS2/SS3 are still out of scope.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- [x] `cargo test --workspace --all-features`
- [x] New cases cover the issue repro (`ESC * 0 ESC N l|` / `ESC + 0 ESC O l|`) and assert `|` stays untranslated so a stuck shift would fail

Signed-off-by on the commit (DCO).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,7 +53,8 @@ listed under a **Changed** or **Removed** heading.
   `ESC N` (SS2) / `ESC O` (SS3) invoke that set for exactly one character
   before the locking shift resumes. The designation was already consumed;
   the shift did nothing, so a mixed line of text and box-drawing showed
-  the letter. A pending single shift does not survive `RIS`. Locking
+  the letter. A pending single shift is consumed by the next character —
+  including a multi-byte UTF-8 one — and does not survive `RIS`. Locking
   shifts remain G0/G1 only (`SO`/`SI`); `LS2`/`LS3` and `DECSC`/`DECRC` of
   charset state are still unmodelled. (#235)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,16 @@ listed under a **Changed** or **Removed** heading.
 
 ### Fixed
 
+- **G2/G3 designation and SS2/SS3 single shifts are modelled, so a
+  one-character line-drawing shift draws `┌` rather than `l`.** `ESC * 0`
+  / `ESC + 0` designate the DEC Special Graphics set into G2/G3, and
+  `ESC N` (SS2) / `ESC O` (SS3) invoke that set for exactly one character
+  before the locking shift resumes. The designation was already consumed;
+  the shift did nothing, so a mixed line of text and box-drawing showed
+  the letter. A pending single shift does not survive `RIS`. Locking
+  shifts remain G0/G1 only (`SO`/`SI`); `LS2`/`LS3` and `DECSC`/`DECRC` of
+  charset state are still unmodelled. (#235)
+
 - **`find` no longer matches the blank padding past the end of a row.** Its
   single-row path searched the row padded out to the terminal width while
   `contains` searched the trimmed text, so `find("Total: ")` was `Some` on
--- a/README.md
+++ b/README.md
@@ -226,12 +226,13 @@ design. termlens's position:
   reflowed** by a `resize`: rows keep the width they were captured at, by
   decision (`Terminal::resize` says why). The visible grid stays the
   fully-featured surface.
-- **Character sets: G0 and G1 only, and one set translated.** `ESC ( Ps` /
-  `ESC ) Ps` designations and the `SO`/`SI` locking shifts are modelled, and
-  the DEC Special Graphics set (`0`) is translated; every other designation —
-  the UK set, the alternate ROMs — reads as ASCII, and G2/G3 with their
-  single shifts are not modelled. `DECSC`/`DECRC` do not save or restore the
-  charset state.
+- **Character sets: G0–G3 designation, SO/SI locking shifts, SS2/SS3
+  single shifts, and one set translated.** `ESC ( ) * + Ps` designations,
+  the `SO`/`SI` locking shifts, and `ESC N`/`ESC O` (SS2/SS3, one character)
+  are modelled, and the DEC Special Graphics set (`0`) is translated; every
+  other designation — the UK set, the alternate ROMs — reads as ASCII.
+  Locking shifts remain G0/G1 only (`LS2`/`LS3` are not modelled).
+  `DECSC`/`DECRC` do not save or restore the charset state.
 - `wait_frame` needs the application to bracket its repaints in DEC 2026
   synchronized updates, and only the last 8 completed frames are retained;
   everything else waits with `wait_until`, under the three rules in
--- a/crates/termlens/src/emu/seq.rs
+++ b/crates/termlens/src/emu/seq.rs
@@ -166,7 +166,7 @@ fn dec_special_graphics(b: u8) -> Option<&'static str> {
     })
 }
 
-/// SS2/SS3 invoke G2/G3 for exactly one GL character, then the locking
+/// SS2/SS3 invoke G2/G3 for exactly one character, then the locking
 /// shift resumes. Stored as which *slot* to read, not a copy of the set:
 /// a designation between the shift and the character still applies.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -399,8 +399,8 @@ pub(crate) struct SeqTracker {
     /// True after `SO` (`0x0e`) invoked G1; `SI` (`0x0f`) returns to G0.
     /// Single shifts override this for one character without changing it.
     shifted_out: bool,
-    /// Pending SS2/SS3, if any. Cleared by the next GL character, or by
-    /// RIS; an intervening control or designation must not consume it, or a
+    /// Pending SS2/SS3, if any. Cleared by the next character, or by RIS;
+    /// an intervening control or designation must not consume it, or a
     /// mixed line that shifts then redraws would lose the graphic.
     single_shift: Option<SingleShift>,
     /// Which introducer opened the current DCS-class string: `P` (DCS),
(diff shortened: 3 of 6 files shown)


vyncint (changes requested):
Thank you — this is a careful piece of work on the hardest of the batch, and it lands the parts I expected to be got wrong. The one-character lifetime is tested from both directions (`\033*0\033Nll` → `┌l`), the shift correctly overrides `SO` without leaving G1, `RIS` clears a pending shift, and you found the double-read problem in `graphics_glyph` on your own and documented the fix in `docs/DESIGN.md`. Using `|` as the sentinel because it is *itself* a Special Graphics byte (`≠`) is the right instinct: a stuck shift shows up as a wrong glyph rather than as a missing one.

Verified locally against a build of the branch — the issue's repro, the locking-shift interaction, a `CSI` and an `OSC` between the shift and the character, two `SS2` in a row, `SS2` at end of stream, and `SS2` then `SI` then the character. All behave as you describe. `fmt`, `clippy --all-targets --all-features -D warnings` and the full suite are clean here too.

One blocking problem.

## A pending shift survives a multi-byte character and translates the next one

The shift is consumed on `0x20..=0x7e`, so any character outside that range does not consume it. Measured on `e1a609c`:

```
$ printf '\033*0\033N\346\261\211l'     # SS2, then 汉, then l
汉┌                                      # want: 汉l
$ printf '\033*0\033N\360\237\246\200l' # SS2, then 🦀, then l
🦀┌                                      # want: 🦀l
$ printf '\033*0\033N\303\251l'         # SS2, then é, then l
é┌                                       # want: él
$ printf '\033*0\346\261\211l'          # control: same bytes, no SS2
汉l                                      # correct, so the shift is the cause
```

`SS2` applies to the next *character*. `汉` is that character — it has no Special Graphics mapping, so it renders as itself and the shift is spent. Instead the shift stays pending across the three continuation bytes and lands on the `l` after it, which becomes `┌`.

This is the failure mode the issue exists to remove, arriving one character later: a letter silently becomes box drawing, a test asserting the letter fails against an application that is correct, and a snapshot that blessed the glyph keeps passing. A mixed line of CJK or emoji and box drawing is exactly the case a single shift is for, and this crate carries a `unicode-torture` fixture precisely because that traffic is expected.

I take the PR description's "consumed on the next GL character" as the intent rather than a scope limit, because nothing in the code or the docs tells a reader that a multi-byte character is exempt, and the consequence is a corrupted *following* character rather than a missing feature.

**Suggested fix**, at the consumption site in `crates/termlens/src/emu/seq.rs`. A single shift lasts one character, so consume at a character *start* and let continuation bytes alone:

```rust
// A single shift lasts for one *character*, not one byte: a multi-byte
// UTF-8 character consumes it too, so a continuation byte must not, or
// the shift survives `汉` and translates the byte after it.
let utf8_continuation = (0x80..=0xbf).contains(&b);
if b >= 0x20 && b != 0x7f && !utf8_continuation {
    self.single_shift = None;
}
```

I applied exactly this to your branch to check it: the three cases above become `汉l`, `🦀l`, `él`; all 194 lib tests, `charset`, `fixtures`, `styles` and `scrollback` stay green; and every behaviour this PR claims is unchanged, including your four new tests.

Two things I checked and would **not** change, so they do not get swept up:

- **`DEL` (0x7f) does not consume the shift** — `\033*0\033N\177l` gives `┌`. `DEL` is a control that terminals ignore, so the shift correctly falls through to `l`. The `b != 0x7f` above keeps that.
- **The shift survives a newline** — `\033*0\033N\nl` puts `┌` at the start of the next row. A newline is C0, and the shift applies to the next graphic character wherever it lands. Defensible, and consistent with your "intervening C0 must not consume it" rule.

## Smaller, non-blocking

The module docs in `seq.rs` and the `SingleShift` doc comment both say "one GL character". Once the fix above lands, "one character" is the accurate wording — worth changing so the next reader does not reintroduce the byte-range test from the comment.

Please add a case for this to `crates/termlens/tests/charset.rs` beside the ones you wrote — a CJK or emoji character between the shift and a graphics byte, asserting the byte after it is *not* translated. That is the assertion that would have caught this, and it belongs next to the `|` sentinel trick for the same reason.

Happy to re-review as soon as it is pushed.

cestercian:
Thanks — good catch on the multi-byte path.

Pushed `de6b7f1`: consume SS2/SS3 at a character start (`b >= 0x20 && b != 0x7f && !utf8_continuation`), so `汉` / `🦀` / `é` spend the shift and the following `l` stays a letter. DEL and C0 still do not consume it. Docs now say “one character”; added unit + PTY coverage beside the `|` sentinel tests.

Happy to take another look whenever you have a moment.

vyncint (approved):
Thanks — this is exactly the follow-up I hoped for, and it went further than the request.

The fix is the right one and I re-verified it against a build of `de6b7f1`:

```
printf '\033*0\033N\346\261\211l'        → 汉l     (was 汉┌)
printf '\033*0\033N\360\237\246\200l'    → 🦀l     (was 🦀┌)
printf '\033*0\033N\303\251l'            → él      (was é┌)
```

and re-ran every probe from the first pass, all still correct: the one-character lifetime (`┌|` and `┌l`), SS3 via `ESC + 0` / `ESC O`, a `CSI` and an `OSC` between the shift and the character, `SO` + SS2 giving `l─┐` without leaving G1, `RIS` clearing a pending shift, two `SS2` in a row, and `SS2` at end of stream. A CJK character *twice* before the graphics byte also behaves (`汉汉l`), which is the case a naive "clear on the next byte ≥ 0x80" would have broken.

Three things I want to call out, because they are the difference between a fix and a fix that stays fixed:

- **Both new tests genuinely fail without the change.** I reverted only the consumption rule on your branch and confirmed the unit test reports `shift must not survive 汉 and translate the next l` and the integration test reports `left: "汉┌"  right: "汉l"`, then restored it and watched both go green. A test that passes either way would have been worse than none, and these are real guards.
- **You pinned the two behaviours I asked you to leave alone** rather than tidying them away — `\x1b*0\x1bN\x7fl` → `┌` is now asserted in `a_single_shift_survives_intervening_controls_and_designations`, and the comment records that C0, `DEL` and newline do not consume the shift. That is the part that stops someone "simplifying" this back into a byte-range test next year.
- Reaching for the tracker directly in the unit test because `drawn` skips UTF-8, instead of bending the helper, was the right call.

Docs are consistent too: the module docs, the `SingleShift` comment, the `SS2`/`SS3` comment, `docs/DESIGN.md` and the CHANGELOG all now say "one character" rather than "one GL character", and the DESIGN scope paragraph still correctly lists `LS2`/`LS3` and `DECSC`/`DECRC` of charset state as unmodelled — the latter is #232, still open, and this PR deliberately does not touch it.

Verified locally on `de6b7f1`: `fmt`; clippy under all four feature configurations; the full suite under `--all-features`, default, `--no-default-features` and `--no-default-features --features decode`; the release profile; `RUSTDOCFLAGS='-D warnings' cargo doc`; and `cargo +1.85 check --workspace --locked --all-targets`. All clean.

One unrelated note so it is on the record and not mistaken for something this PR did: `builder_validation::the_default_working_directory_is_the_test_process_s` fails in a checkout whose absolute path is longer than about 80 characters, because it reads `pwd`'s output from row 0 of an 80-column screen and a longer path wraps. I reproduced it on `main` at `5394c11` in both profiles, and it passes in a short path — so it predates this branch. That is my bug from #228; I am filing it separately.

Merging. Thank you for taking the hardest one in the batch.

PR #267, opened 2026-09-05
docs: label feature-gated APIs on docs.rs

### LINKED ISSUE #258: docs.rs shows the `decode` and `insta` items with no feature badge, so optional API reads as unconditional
**Today** — everything behind the `decode` feature — `Bitmap` and its methods, `DecodeError`, and `GraphicsPayload::decode` — plus the `insta` re-export and `assert_screen_snapshot!` behind the `insta` feature, appears on docs.rs with no indication that any of it is optional. `crates/termlens/Cargo.toml:14` asks docs.rs to build with everything on:

```toml
[package.metadata.docs.rs]
all-features = true
```

but nothing tells rustdoc to *label* what those features gate:

```
$ grep -rn 'docsrs\|doc(cfg' crates/termlens/src/
$                                    # nothing
```

So `Bitmap`, `DecodeError`, `GraphicsPayload::decode` and the rest are rendered exactly like `Screen::cell` — as part of the crate. A reader following the README (`cargo add termlens --dev`, which does **not** enable `decode`) writes `payload.decode()?`, and finds out from rustc that the method does not exist. The docs gave no warning, and the item they read is on the page they were sent to.

**Why it is worth fixing** — this crate's documentation is its contract, and the one thing a reader cannot see in it is which half of it they actually installed. `decode` is off by default deliberately, and the README says to add `--features decode` "if you test an application that draws inline images" — one sentence, three screens above the API reference where the decision actually bites. The standard remedy exists, is one line of manifest plus one attribute per gated item, and produces the familiar "Available on crate feature `decode` only" badge that every reader of Rust docs already knows how to read.

**Fix** — the conventional pattern:

```toml
[package.metadata.docs.rs]
all-features = true
rustdoc-args = ["--cfg", "docsrs"]
```

```rust
// lib.rs, at the top
#![cfg_attr(docsrs, feature(doc_cfg))]
```

then on each gated item, beside its existing `#[cfg(...)]`:

```rust
#[cfg(feature = "decode")]
#[cfg_attr(docsrs, doc(cfg(feature = "decode")))]
pub fn decode(&self) -> Result<Bitmap, DecodeError> { … }
```

`doc_cfg` is nightly-only, which is why it is behind the `docsrs` cfg that only docs.rs sets — stable builds and the MSRV job never see the attribute, and the local `cargo doc` gates in CONTRIBUTING keep working unchanged. Check the result locally with:

```sh
cargo +nightly rustdoc -p termlens --all-features -- --cfg docsrs
```

Worth doing in the same pass: the `insta` re-export and `assert_screen_snapshot!` are gated on the default-on `insta` feature, so they need the badge too — a consumer who turned default features off is in exactly the same position.

**Done when** — every `#[cfg(feature = …)]` public item carries the matching `doc(cfg)`; the manifest passes `--cfg docsrs` to docs.rs; a nightly `cargo rustdoc --cfg docsrs` shows the feature badge on `Bitmap` and on `assert_screen_snapshot!`; and the stable `cargo doc` gates in CI are unaffected.

nightcityblade:
## What & why

Configure docs.rs to expose feature-gated API badges, and annotate the public `decode` and `insta` APIs with their required features.

Closes #258.

## Checklist

- [x] Tests added/updated for the change (documentation output verified directly)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,11 @@ listed under a **Changed** or **Removed** heading.
   cell and style assertions. Every Rust block in it is compiled against the
   crate in CI, and the README shows the one-line install for Claude Code.
 
+### Fixed
+
+- **docs.rs labels APIs gated by the `decode` and `insta` features.** Optional
+  items no longer appear to be available in every build. (#258)
+
 ## [0.9.0] - 2026-09-05
 
 ### Added
--- a/crates/termlens/Cargo.toml
+++ b/crates/termlens/Cargo.toml
@@ -13,6 +13,7 @@ authors.workspace = true
 
 [package.metadata.docs.rs]
 all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
 
 [features]
 default = ["insta"]
--- a/crates/termlens/src/graphics.rs
+++ b/crates/termlens/src/graphics.rs
@@ -301,6 +301,7 @@ impl GraphicsPayload {
     /// size. [`DecodeError::TooLarge`] covers declared dimensions or sixel
     /// commands beyond the 4096-pixel and related safety limits.
     #[cfg(feature = "decode")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "decode")))]
     pub fn decode(&self) -> Result<Bitmap, DecodeError> {
         let data = self.data.as_deref().ok_or(DecodeError::NotCaptured)?;
         match self.protocol {
@@ -386,6 +387,7 @@ impl GraphicsPayload {
 /// be about the picture — "the day at week 30 is Primer's brightest green" —
 /// rather than about its size in bytes.
 #[cfg(feature = "decode")]
+#[cfg_attr(docsrs, doc(cfg(feature = "decode")))]
 #[derive(Clone, PartialEq, Eq)]
 pub struct Bitmap {
     width: u32,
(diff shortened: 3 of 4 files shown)


vyncint (changes requested):
Reviewed and verified. This is correct and complete.

What I checked beyond CI:

- Built the docs on nightly the way docs.rs will (`cargo +nightly rustdoc -p termlens --all-features -- --cfg docsrs`) and confirmed the badges render: `Bitmap`, `DecodeError` and `GraphicsPayload::decode` show "Available on crate feature `decode` only", and `assert_screen_snapshot!` shows the `insta` one.
- Confirmed nothing changes on stable: `docsrs` is never set by a normal build, so `feature(doc_cfg)` stays inert and clippy with `-D warnings` is clean in all four feature configurations. Cargo already knows `docsrs` as a well-known cfg, so there is no `unexpected_cfgs` warning to suppress.
- Checked the coverage: `decode` gates exactly three public declarations plus the re-export, and `insta` gates the re-export and the macro. You annotated all of them. `Bitmap`'s own methods sit inside a gated `impl`, so they inherit the type's badge and need nothing.

A nice side effect worth knowing about: enabling `doc_cfg` also makes rustdoc label the `#[cfg(unix)]` items automatically, so `Signal` and `Terminal::signal` now carry "Available on Unix only" without either of us writing an attribute for them. That is an improvement, not a problem.

---

```sh
```

nightcityblade:
Rebased onto current `main` and moved the docs.rs entry under `Unreleased` → `Fixed` in 8dc0380.

vyncint (approved):
Thanks — all three points are addressed, and the entry is under `[Unreleased]` now.

Re-verified on this head: badges render on nightly with `--cfg docsrs` for `Bitmap`, `DecodeError`, `GraphicsPayload::decode` and `assert_screen_snapshot!`; stable stays clean in all four feature configurations; docs.rs metadata is right. Merging.

PR #268, opened 2026-09-05
fix: reject out-of-bounds row text access

### LINKED ISSUE #260: `row_text` answers an out-of-range row with `""`, the quiet answer `rect_text` refuses to give
**Today** — `Screen::row_text` returns an empty string for a row that does not exist, and says so in its own rustdoc. Measured against 0.9.0 on a 24-row screen:

```rust
let s = t.screen();
s.row_text(99)          // ""   — no row 99; indistinguishable from a blank row
s.cell(99, 0)           // None — the same question, answered
s.rect_text(0.., 99..)  // ""   — clamped, deliberately
```

Three neighbouring accessors, three different answers to "you asked for something outside the grid".

**Why it is worth fixing** — the crate has already decided this question once, in the opposite direction, and wrote the reasoning down. `rect_text` panics on a *backwards* range, and its rustdoc explains why at length (screen.rs, the "Panics" section):

> A panic rather than an error, deliberately, and for the same reason `&slice[3..0]` panics: a backwards range is not a fact about the terminal discovered at runtime, it is a mistake in the calling source. Returned quietly it read as `""` — "this pane is empty", a perfectly plausible assertion outcome — so a mis-ordered call passed for the wrong reason and kept passing.

That paragraph describes `row_text(99)` exactly. `assert_eq!(s.row_text(30).trim(), "")` on a 24-row screen passes, and passes for the wrong reason, and keeps passing after the application stops drawing anything at all. An off-by-one over `rows()`, a hardcoded row index that outlived a `size()` change, or a `(col, row)` pair passed row-first are all the same mistake in the calling source, and all of them are answered `""`.

The documented behaviour is not a defence: it is documented as a fact, without the reasoning `rect_text` gives for the opposite choice, and a reader who has read `rect_text` would reasonably expect the other answer.

**Fix** — pick one of the two answers the crate already gives, and make the trio consistent. In order of preference:

1. **Panic**, like `rect_text`, with a message in the same shape: `row_text: row 99 is outside the 24-row screen`. It is the same class of mistake and gets the same treatment. Breaking only for callers who were relying on the quiet `""`, which is the bug being fixed.
2. **Return `Option<String>`**, like `cell`. More honest still, and the most disruptive: every call site grows an `.unwrap()` or a `?`, including the ones in this repository's own suite.

Whichever is chosen, `text()` and `rect_text` must keep working — `text()` iterates `0..self.rows` so it never asks out of range, and `rect_text` clamps before it reads. Note the distinction `rect_text` draws and keep it: an out-of-*range* bound is clamped because "asking for more screen than exists is a reasonable thing to do", while an inverted range is a mistake. A single row index has no range to clamp, so it falls on the mistake side.

Search the suite for `row_text(` before changing anything: the calls there all index within the grid, so the migration should be small, and any that do not are themselves the bug this finds.

**Done when** — `row_text` no longer answers an out-of-range row with `""`; its rustdoc carries the reasoning rather than only the behaviour, and cross-references `cell` and `rect_text` so the trio reads as one decision; a test covers the out-of-range call; and `text()` and `rect_text` are unchanged.

nightcityblade:
## What & why

Make `Screen::row_text` panic with a descriptive message for an out-of-bounds row, preventing invalid coordinates from looking like a legitimately blank row. The docs now distinguish this behavior from `cell` and clamped `rect_text` ranges, and a focused regression test covers it.

Closes #260.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,13 @@ listed under a **Changed** or **Removed** heading.
   cell and style assertions. Every Rust block in it is compiled against the
   crate in CI, and the README shows the one-line install for Claude Code.
 
+### Changed
+
+- **`Screen::row_text` rejects an out-of-bounds row instead of returning an
+  ambiguous empty string.** Callers with a potentially invalid index should
+  bounds-check against `rows()` first, or use `cell(row, 0)` and treat `None`
+  as absent. (#260)
+
 ### Fixed
 
 - **docs.rs labels APIs gated by the `decode` and `insta` features.** Optional
--- a/crates/termlens/src/screen.rs
+++ b/crates/termlens/src/screen.rs
@@ -855,19 +855,29 @@ impl Screen {
     }
 
     /// The text of one row, blank cells rendered as spaces, trailing
-    /// whitespace **included**. Returns an empty string for out-of-bounds
-    /// rows.
+    /// whitespace **included**.
     ///
     /// Wide characters contribute their character once; their continuation
     /// cell contributes nothing (so the string's *display width* matches the
     /// row, not its `char` count).
+    ///
+    /// # Panics
+    ///
+    /// Panics when `row` is outside the screen. Unlike the clamped ranges
+    /// accepted by [`rect_text`](Self::rect_text), a single invalid index is
+    /// a caller mistake; returning an empty string would make it
+    /// indistinguishable from a blank row. Use [`cell`](Self::cell) when an
+    /// out-of-bounds result is expected and should be represented by `None`.
     #[must_use]
     pub fn row_text(&self, row: u16) -> String {
+        assert!(
+            row < self.rows,
+            "row_text: row {row} is outside the {}-row screen",
+            self.rows
+        );
         let mut out = String::with_capacity(usize::from(self.cols));
         for col in 0..self.cols {
-            let Some(cell) = self.cell(row, col) else {
-                return out;
-            };
+            let cell = self.cell(row, col).expect("row and column are in bounds");
             if cell.is_wide_continuation() {
                 continue;
             }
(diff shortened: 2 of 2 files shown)


vyncint (changes requested):
Reviewed and verified. The implementation is right.

What I checked beyond CI:

- The panic boundary is exact: on a 20x4 screen `row_text(3)` returns its row and `row_text(4)` panics with `row_text: row 4 is outside the 4-row screen`. The message follows `rect_text`'s shape, which is what makes the two read as one decision.
- The only internal caller is `text()`, which iterates `0..self.rows`, so nothing inside the crate can trip the new assertion. Full suite green, both profiles.
- The `expect` replacing the `let … else` is unreachable by construction once the assertion has run, and it keeps the branch that was already there rather than adding one.

The `# Panics` section reads well, and pointing at `cell` for the case where `None` is the wanted answer is the right cross-reference.

### One more change, alongside the shared one below

The entry belongs under **Changed**, not **Fixed**. The CHANGELOG preamble states the rule:

> Until 1.0, minor versions (0.x) may contain breaking changes; they are always listed under a **Changed** or **Removed** heading.

A call that used to return `""` and now panics is a behaviour a user must relearn, so it needs to be where someone scanning for breakage will look. Worth adding the one-line migration too, in the shape the neighbouring entries use: a caller who was reading a row index that may be out of range should bounds-check against `rows()` first, or use `cell(row, 0)` and treat `None` as absent.

---

```sh
```

nightcityblade:
Rebased onto current `main` and updated the changelog in 30702e2: the entry now sits under `Unreleased` → `Changed` and includes the requested migration guidance.

vyncint (approved):
Thanks — both points addressed: the entry moved to `[Unreleased]` under **Changed**, and the migration line tells a caller with a possibly-invalid index what to do instead.

Re-verified on this head: the panic boundary is exact (`row_text(3)` returns its row on a 4-row screen, `row_text(4)` panics with the message naming both numbers), `text()` is unaffected because it iterates `0..rows`, and the full suite is green.

I rebased the branch onto `main` myself and resolved the changelog collision with #267, which landed a few minutes ago and claimed the same lines. Your commit, author and sign-off are unchanged; the only difference is that `[Unreleased]` now carries **Added**, **Changed** and **Fixed** in Keep a Changelog order. Nothing for you to do. Merging.

PR #269, opened 2026-09-05
fix: make drag coordinate order explicit

### LINKED ISSUE #257: `drag` takes (col, row) tuples, so a `find` result passed straight in transposes the gesture silently
**Today** — `Terminal::drag` takes its endpoints as `(u16, u16)` tuples in **column-first** order, and `Screen::find` returns `(u16, u16)` in **row-first** order. They are the same type, so feeding one to the other compiles and runs, silently transposed. Measured against 0.9.0:

```rust
// terminal.rs:2543
pub fn drag(&mut self, button: impl Into<MouseChord>, from: (u16, u16), to: (u16, u16)) -> Result<()>

let at = screen.find("Drag me").unwrap();   // (row, col)
t.drag(MouseButton::Left, at, elsewhere)?;  // read as (col, row) — compiles, drags the wrong cell
```

Nothing catches it. The endpoints are bounds-checked against the grid, so a transposed pair inside a squarish grid passes that check too, and the application under test receives a well-formed drag across cells nobody asked about.

Its two siblings cannot be misused this way. `click(col, row)` and `click_with(button, col, row)` take separate arguments (terminal.rs:2471, 2495), so a `find` result has to be destructured before it can be passed, and the author sees the order while doing it. `drag` is the only mouse method that accepts a bare pair.

`drag`'s own rustdoc — forty lines on tracking modes, path interpolation and what each mode delivers — never states the order. The statement lives two files away, in `Screen::rect_text`'s "Panics" section, which lists every column-first API by name (`drag` among them) and says outright that "swapping the two is therefore the mistake to expect". That was written for #219, about exactly this hazard.

**Why it is worth fixing** — every other coordinate hazard in this crate is caught by a type, a bounds check, or a panic. This one is caught by nothing, and it is the single place where the crate's own most convenient source of coordinates (`find`) flows into an input method with the axes reversed. The failure is not a compile error, not a runtime error, and not visibly wrong on the screen — the application simply receives a gesture somewhere else, and the test asserts on whatever that produced.

**Fix** — decide between the two shapes and say why in the PR. Both are small:

1. **Match the siblings:** `drag(button, from_col, from_row, to_col, to_row)`. Four `u16`s is not pretty, but it makes the transposition unwritable and puts `drag` in line with `click`, `click_with`, `scroll` and `scroll_with`. A breaking change to one method, which is a minor bump before 1.0.
2. **Keep the tuples and make the order visible**: state "columns first" in the first line of the rustdoc, next to the parameter, rather than only in `rect_text`'s panic section — and add a doctest that names the order.

Option 1 is the recommendation: the docs already tried option 2 once, in the place a reader of `drag` never looks.

Either way, `drag`'s rustdoc should carry the same cross-reference `rect_text` does, and the `Screen::find` docs should say that its `(row, col)` result is not directly usable as a mouse position.

**Done when** — passing a `Screen::find` result straight into `drag` either does not compile or is documented on `drag` itself in its opening lines; the order appears in `drag`'s own rustdoc rather than only in `rect_text`'s; and a test covers a drag whose endpoints came from `find`, so the intended spelling is exercised.

nightcityblade:
## What & why

Change `Terminal::drag` to accept four explicitly ordered column/row arguments, matching the other mouse APIs. This prevents the row-first tuple returned by `Screen::find` from being passed directly and silently transposed. The relevant docs and drag tests now use the unambiguous spelling.

Closes #257.

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,11 +23,6 @@ listed under a **Changed** or **Removed** heading.
 
 ### Changed
 
-- **`Screen::row_text` rejects an out-of-bounds row instead of returning an
-  ambiguous empty string.** Callers with a potentially invalid index should
-  bounds-check against `rows()` first, or use `cell(row, 0)` and treat `None`
-  as absent. (#260)
-
 ### Fixed
 
 - **docs.rs labels APIs gated by the `decode` and `insta` features.** Optional
--- a/crates/termlens/src/screen.rs
+++ b/crates/termlens/src/screen.rs
@@ -1018,6 +1018,8 @@ impl Screen {
 
     /// Locate the first occurrence of `needle` scanning rows top to bottom;
     /// returns the `(row, col)` of its first character.
+    /// Mouse methods such as [`Terminal::drag`](crate::Terminal::drag) take
+    /// columns first, so destructure this pair before using it as mouse input.
     ///
     /// The **visible screen only**, like [`contains`](Self::contains): a
     /// needle that has scrolled into history is not found here, however
--- a/crates/termlens/src/terminal.rs
+++ b/crates/termlens/src/terminal.rs
@@ -2510,8 +2510,12 @@ impl Terminal {
         self.write_input(&bytes, "a mouse click")
     }
 
-    /// Drag from one cell to another: press, a motion report **per cell
-    /// crossed**, release.
+    /// Drag from `(from_col, from_row)` to `(to_col, to_row)`: press, a
+    /// motion report **per cell crossed**, release.
+    ///
+    /// Coordinates are columns first, matching [`click`](Self::click) and
+    /// the other mouse methods. [`Screen::find`](crate::Screen::find) returns
+    /// `(row, col)` instead, so destructure its result before passing it here.
     ///
     /// The path is a straight line, interpolated on both axes for a diagonal
     /// drag. A real pointer's exact path is not reproducible and does not
@@ -2543,8 +2547,10 @@ impl Terminal {
     pub fn drag(
         &mut self,
         button: impl Into<MouseChord>,
-        from: (u16, u16),
-        to: (u16, u16),
+        from_col: u16,
+        from_row: u16,
+        to_col: u16,
+        to_row: u16,
     ) -> Result<()> {
         self.ensure_deliverable("a mouse drag")?;
         let chord = button.into();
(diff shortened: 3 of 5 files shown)


vyncint (changes requested):
Reviewed and verified. The signature change does exactly what the issue asked.

What I checked beyond CI:

- The transposition is now a compile error rather than a silent wrong gesture: passing a `Screen::find` result straight in fails with "this method takes 5 arguments but 3 arguments were supplied".
- The drag suite passes unchanged, including the per-cell motion test and the off-grid endpoint test, so the four arguments reach `check_mouse_in_grid` in the same order the tuples did.
- Reconstructing `from`/`to` inside the body keeps the diff to the boundary, which is the right call.

Taking option 1 from the issue was the better of the two, and the cross-reference you added on `Screen::find` is the half I would have forgotten.

### One more change, alongside the shared one below

`skills/termlens/SKILL.md` landed on `main` a few hours after you opened this, and it documents the old two-tuple signature in three places. It is the file coding agents copy from, so it has to move with the API:

- **line 99**, in the coordinate-order rule: `` `drag(button, (col, row), (col, row))` `` becomes `` `drag(button, from_col, from_row, to_col, to_row)` ``, and the sentence after it that says "the tuple types match and the axes do not" now needs to say that the four arguments make the transposition unwritable.
- **line 369**, in the API cheat sheet: `` `drag(MouseButton::Left, (c, r), (c, r))` ``.
- **line 414**, in the pitfalls table: the row whose left column is `` `t.drag(b, s.find("x").unwrap(), …)` ``.

`.github/scripts/check-skill-snippets.sh` compiles the skill's ```rust blocks, but these three are prose and tables, so CI cannot catch them. Sorry for the collision — it is entirely an accident of merge order, not something you could have seen when you opened this.

---

```sh
```

nightcityblade:
Addressed both follow-ups in f8c78e4: rebased and moved the changelog entry under `Unreleased` → `Changed`, and updated all three skill references to the four-argument `drag` API.

Validation: `cargo fmt --check`, `cargo test --workspace --all-features`, and `.github/scripts/check-skill-snippets.sh` pass.

vyncint (approved):
Thanks — both points addressed, and the skill update is thorough. You caught all three places, and the rewrites are accurate rather than mechanical: the rule now explains what the four arguments buy, and the pitfalls row describes the mistake in words instead of showing a call that no longer compiles.

Re-verified on this head: passing a `Screen::find` result straight into `drag` is a compile error, the drag suite passes unchanged including the per-cell motion and off-grid endpoint cases, and the skill's compile guard is green.

I rebased the branch onto `main` and resolved the changelog collision with #267 and #268, which landed ahead of it. Your commit, author and sign-off are untouched. Merging.

PR #270, opened 2026-09-05
fix: make Style non-exhaustive

### LINKED ISSUE #259: `Style` is the one public struct that is not `#[non_exhaustive]`, and two attributes are already known to be missing
**Today** — `Style` is the only public struct in the crate with public fields, and the only public type without `#[non_exhaustive]`. Measured against 0.9.0:

```
$ # every public struct with pub fields, and whether it is non_exhaustive
screen.rs   Style   non_exhaustive=NO   pub fields: fg, bg, bold, dim, italic,
                                        underline, reverse, blink, conceal, strikethrough
```

Every public enum — `Error`, `Key`, `Color`, `MouseMode`, `CursorShape`, `Graphics`, `Scroll`, `MouseButton`, `GraphicsProtocol`, `GraphicsAction`, `GraphicsFormat`, `DecodeError` — carries the attribute. `Style` does not, so a consumer may write `Style { fg, bg, bold, dim, italic, underline, reverse, blink, conceal, strikethrough }` as a literal and destructure it exhaustively, and adding an eleventh field breaks them.

**Why it is worth fixing** — an eleventh field is already known to be wanted. #184 closed by *documenting* that overline (`SGR 53`) and double underline (`SGR 21`) reach no `Style`, and the README's Known limitations still says so; the natural fix for that is two more `bool`s. `cargo-semver-checks` runs on every release (`release.yml`), so it will correctly refuse the addition as breaking, and the change will cost a version bump it should not have to.

Adding `#[non_exhaustive]` is *itself* a breaking change, so it can only get cheaper by happening sooner. At 0.9.0 the crate is pre-1.0 with a handful of consumers, and a minor bump costs a line in three manifests. After 1.0 it costs a major.

The property being protected is real rather than theoretical: `Style` is a description of what a terminal drew, and terminals keep growing attributes. Underline colour (`SGR 58`) and underline style (curly, dotted, dashed) are the next two after overline, and a backend that models more of them would want to report them.

**Fix** — add the attribute:

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
#[non_exhaustive]
pub struct Style { … }
```

Reading fields is unaffected — `cell.style().bold` keeps working, which is what nearly every test does. What stops compiling outside the crate is constructing a literal and destructuring exhaustively. Both remain possible through `Style::default()`:

```rust
let mut want = Style::default();
want.fg = Color::Indexed(1);
want.bold = true;
assert_eq!(*cell.style(), want);
```

Three things belong in the same PR. Check whether any in-workspace test constructs a `Style` literal and convert it to the `default()` form, so the pattern the docs recommend is the pattern the suite uses. Add a short rustdoc paragraph on `Style` saying attributes get added and how to build one for comparison. And add the same note to the CHANGELOG under `[Unreleased]` as a **Changed** entry, since a consumer building a literal today will need the one-line migration above.

Worth deciding out loud in the PR and either doing or explicitly declining: a `Style::builder()`-shaped helper, or leaving `default()` plus field assignment as the only spelling. Declining is a reasonable answer — two lines is not a hardship — but it should be a decision rather than an omission.

**Done when** — `Style` is `#[non_exhaustive]`; the rustdoc says attributes may be added and shows how to construct one for comparison; any in-workspace literal is converted; the CHANGELOG carries the migration line; and `cargo-semver-checks` reports the change (which is the point — it is the last cheap moment to make it).

nightcityblade:
## What & why

Mark `Style` as non-exhaustive while the crate is pre-1.0, so future terminal attributes can be represented without another breaking change. The API docs and changelog explain how downstream comparisons should construct a style from `Style::default()`.

A dedicated builder is intentionally omitted: default plus field assignment remains concise and avoids adding API surface solely for test expectations.

Closes #259.

## Checklist

- [x] Tests added/updated for the change (existing style integration suite covers field access and comparisons)

### DIFF
--- a/crates/termlens/src/screen.rs
+++ b/crates/termlens/src/screen.rs
@@ -37,7 +37,11 @@ pub enum Color {
 ///
 /// Inspect them per cell via [`Cell::style`], or snapshot them wholesale
 /// with [`Screen::with_styles`] (plain snapshots stay text-only).
+/// More terminal attributes may be added in future releases. To construct a
+/// style for comparison, start from `Style::default()` and assign the fields
+/// relevant to the assertion rather than using a struct literal.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, Default)]
+#[non_exhaustive]
 pub struct Style {
     /// Foreground color.
     pub fg: Color,


vyncint (changes requested):
Reviewed and verified. Correct, and the reasoning in the PR body is the right call.

What I checked beyond CI:

- Built an external consumer crate against this branch. `Style { … }` is refused with `E0639: cannot create non-exhaustive struct using struct expression`, and so is the functional-update form `Style { bold: true, ..Style::default() }` — which is the one people reach for first and the reason the doc paragraph has to name the alternative explicitly. It does.
- The documented migration compiles: `let mut s = Style::default(); s.fg = …; s.bold = true;`.
- No integration test or fixture constructs a `Style` literal, so nothing outside the crate needed converting. The in-crate literals are unaffected by the attribute.

Declining the builder is the right answer and I am glad you said so out loud rather than leaving it unmentioned. Two lines of assignment is not a hardship, and a builder would be API surface that exists only for test expectations.

This is also the last cheap moment to make this change, which is what the issue was really about — after 1.0 it costs a major.

---

```sh
```

nightcityblade:
Rebased onto current `main` and moved the `Style` entry under `Unreleased` → `Changed` in dcc23e9.

vyncint (approved):
Thanks — the entry is under `[Unreleased]` now, and it was already under the right heading.

Re-verified on this head from an external consumer crate: `Style { … }` and the functional-update form `Style { bold: true, ..Style::default() }` are both refused with `E0639`, and the migration your rustdoc recommends compiles. That second case is the one that matters, since it is what people reach for first and the docs name the alternative explicitly.

I rebased onto `main` and resolved the changelog collision with the three that landed ahead of this. Your commit, author and sign-off are untouched. Merging.

PR #271, opened 2026-09-05
fix: clear inspect child environment by default

### LINKED ISSUE #263: `inspect` inherits your shell's environment, so it shows a screen your test will never see
**Today** — `inspect` spawns the program under test with the environment of whoever ran it, while the suite it exists to debug almost always spawns with `env_clear()`. `crates/termlens/examples/inspect.rs:111`:

```rust
let mut t = match Terminal::builder()
    .size(size.0, size.1)
    .timeout(timeout)
    .args(args)
    .spawn(&program)
```

No `.env_clear()`, and no flag for one. The README's example, the crate docs, `bin!`, and every fixture test in this repository use `env_clear()`, because a hermetic environment is the whole point of a harness.

So the tool whose job is "show me what my app looks like in the grid" shows a grid built under different conditions from the one the test will see. Anything the application reads from the environment can differ: `NO_COLOR`, `COLORTERM`, `LANG` and `LC_ALL` (which decide whether it draws UTF-8 box characters or ASCII), `TERM` if the caller exported one, and every application-specific variable a developer happens to have set in their shell.

**Why it is worth fixing** — the failure this produces is the worst kind for a debugging tool: it disagrees with the thing being debugged, and it disagrees quietly. A developer whose snapshot test fails runs `inspect` to see what the application actually drew, gets a *different* screen from the one in the failure, and now has two mysteries. An application that draws a coloured border under the developer's `COLORTERM=truecolor` and a plain one under the test's cleared environment is exactly the case someone reaches for `inspect` to understand.

It is also the cheapest possible fix — one builder call and one flag — for a tool that is the crate's shop window: it is what the README points a new user at, and what `docs/HANDOFF.md` recommends for seeing "any program through termlens's eyes".

**Fix** — make the default match what a test does, and keep the escape hatch:

- `inspect` calls `.env_clear()` by default, so its screen is the screen a test gets.
- `--inherit-env` restores the current behaviour, for the case where the program genuinely needs something from the caller's environment to run at all.
- `--env KEY=VALUE`, repeatable, so a single variable can be set without giving up hermeticity — this is what a developer actually wants when chasing "does it look different with `NO_COLOR=1`?", which is a good enough reason on its own.

Two things to handle while there. Under `env_clear()` the crate refuses a bare program name, because clearing the environment removes `PATH` (#222) — so `inspect ls` would start failing with the error #222 added. Either pass `PATH` through by default (a documented exception, the way `TERM` and `SHELL` already are) or catch that error and add one line telling the user about `--inherit-env`; say which was chosen and why. And the usage text and the file's `//!` header both list the flags, so both need the new ones — they are already the single source `--help` prints, so this is one edit.

**Done when** — `inspect` clears the environment by default and its trailer or usage says so; `--inherit-env` and a repeatable `--env KEY=VALUE` exist; a bare program name still works or fails with a message naming the flag that fixes it; the doc comment and usage text list every flag; and `crates/termlens/tests/inspect.rs` covers the cleared default by inspecting a program that prints its own environment.

nightcityblade:
## What & why

Make the `inspect` example clear the child environment by default so it matches hermetic termlens tests. `PATH` is deliberately retained so bare program names still resolve; repeatable `--env KEY=VALUE` adds selected variables, while `--inherit-env` restores the previous behavior. End-to-end coverage verifies both clearing and selective injection.

Closes #263.

## Checklist

### DIFF
--- a/crates/termlens/examples/inspect.rs
+++ b/crates/termlens/examples/inspect.rs
@@ -5,6 +5,7 @@
 //! cargo run --example inspect -- ls -la
 //! cargo run --example inspect -- --size 120x40 htop
 //! cargo run --example inspect -- --timeout 30 ./target/debug/slow-app
+//! cargo run --example inspect -- --env NO_COLOR=1 my-app
 //! ```
 //!
 //! Waits for the program to exit (up to the deadline, `--timeout`, five
@@ -33,12 +34,15 @@ use termlens::Terminal;
 /// The one copy of the usage text: `--help` prints it to stdout and exits
 /// 0, a missing program prints it to stderr and exits 1 (#229).
 const USAGE: &str = "\
-usage: inspect [--size COLSxROWS] [--timeout SECONDS] [--idle MILLIS] <program> [args…]
+usage: inspect [--size COLSxROWS] [--timeout SECONDS] [--idle MILLIS]
+               [--inherit-env] [--env KEY=VALUE]... <program> [args…]
 
 Runs <program> in an 80x24 pseudo-terminal (or --size), waits for it to
 exit or for the deadline (--timeout, default 5 seconds), and prints the
 rendered screen. A program still running at the deadline is snapshotted
 after --idle milliseconds (default 300) of output silence, then killed.
+The child environment is cleared by default except for PATH; --inherit-env
+keeps the caller's environment, and repeatable --env sets selected values.
 
 Exit code 0: a screen was printed; the trailer under it says what the
 program did. Exit code 1: inspect itself could not run — bad arguments,
@@ -71,6 +75,8 @@ fn main() -> ExitCode {
     let mut size = (80u16, 24u16);
     let mut timeout = Duration::from_secs(5);
     let mut idle = Duration::from_millis(300);
+    let mut inherit_env = false;
+    let mut env = Vec::new();
 
     // Options come before the program; everything after it is the
     // program's own, however flag-like it looks.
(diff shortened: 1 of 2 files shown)


vyncint (changes requested):
Reviewed and verified. The behaviour is right and the `PATH` decision is the one I would have made.

What I checked beyond CI, by running the built example:

- Default clears the environment: a `FOO=leaked` in the caller's shell reaches the child as unset, while `PATH` survives.
- `--env` is genuinely repeatable (`--env A=1 --env B=2` sets both), splits on the first `=` so `--env 'K=a=b'` gives `K` the value `a=b`, and rejects both a missing `=` and an empty key in one line with exit 1, in the same shape `--size` uses.
- `--inherit-env` restores the old behaviour exactly.
- A bare program name still resolves, which is the whole point of keeping `PATH` — and your new test exercises that path by spawning `sh`, which is neat.
- `--help` lists both new flags, and the usage text stays one source.

Keeping `PATH` is a documented exception rather than a hole, and the usage text says so, which is what the issue asked for.

### Two more changes, alongside the shared one below

1. **The new test's assertion has no failure message.** Every other assertion in `crates/termlens/tests/inspect.rs` prints the output it was reading — `"terminal size missing from:\n{stdout}"` and so on — so a failure shows what the screen actually held. As written, a regression here prints `assertion failed: …contains("unset:yes")` and nothing else. Please give it the same treatment.

2. **`--inherit-env` has no test.** It is the flag that restores the previous behaviour, so it is the one most likely to be quietly broken by a later change to the argument loop. One more case in the same test, asserting that a variable set on the `Command` does reach the child under `--inherit-env`, would close it. That also makes the pair of cases read as one statement about the policy.

---

```sh
```

nightcityblade:
Addressed all requested changes in 5d3aec5: rebased and moved the changelog entry under `Unreleased` → `Changed`, added assertion diagnostics, and added coverage for `--inherit-env`.

vyncint (approved):
Thanks — all three points addressed. Both assertions now print what they read, and the `--inherit-env` case is covered, which is the flag most likely to be broken quietly by a later change to the argument loop.

Re-verified on this head by running the built example: the default clears the environment while keeping `PATH`, `--env` is repeatable and splits on the first `=`, a malformed value is one line and exit 1, `--inherit-env` restores the old behaviour, and a bare program name still resolves.

I rebased onto `main` and resolved the changelog collision with the three that landed ahead of this. Your commit, author and sign-off are untouched. Merging — and thank you for five well-scoped PRs in a day, each with the reasoning written down.

PR #273, opened 2026-09-05
fix: honour custom tab stops (HTS, TBC, CHT, CBT)

### LINKED ISSUE #262: Custom tab stops are ignored: `HTS`, `TBC`, `CHT` and `CBT` all do nothing
**Today** — tab stops are fixed at every eighth column. Setting one (`HTS`, `ESC H`), clearing one (`TBC`, `CSI g`) and moving between them (`CHT`, `CSI I`; `CBT`, `CSI Z`) are all parsed and ignored. Measured against 0.9.0 on a 24-column terminal, with `·` marking where the text lands:

```
$ inspect --size 24x2 sh -c "printf 'a\011b'"                        # default stops
a·······b                                                            # column 8, correct

$ inspect --size 24x2 sh -c "printf '\033[4G\033H\033[1Ga\011b'"     # set a stop at column 4, then tab to it
a·······b                                                            # xterm: a··b, at column 4

$ inspect --size 24x2 sh -c "printf '\011\011X\033[1Zy'"             # two tabs, then one back-tab
················Xy                                                   # xterm: y at column 8
```

The default stops are right because they are the backend's hardcoded eight; nothing else about tab stops exists. The escape sequences reach a dispatch table that has no entry for them and vanish.

**Why it is worth fixing** — all four capabilities are in the terminfo entry termlens hands every child by default:

```
$ infocmp -1 xterm-256color | grep -E '^\s+(hts|tbc|cbt)='
	cbt=\E[Z,
	hts=\EH,
	tbc=\E[3g,
```

An application that lays a table out by setting its own stops and tabbing between them — which is what the capabilities are for, and what ncurses uses when a program calls them — renders with every column in the wrong place. The failure is a silent misalignment: the characters are all present, so `contains` still passes and only a column assertion or a whole-screen snapshot catches it, and a snapshot taken before the tab stops were set will happily bless the wrong layout.

Back-tab (`CBT`) has a second life worth naming: it is what `Shift-Tab` sends, and an application echoing or forwarding it emits `CSI Z` on its *output* side. Today that moves nothing.

**Fix** — the backend has no notion of tab stops beyond its fixed eight, so the state belongs where the other state it does not model already lives: `emu/seq.rs`, beside the character sets, focus reporting and the cursor shape.

A tab stop set is a bitmap one bit per column, `cols` wide, initialised to every eighth column and reset there by `RIS` — and `DECSTR` too, whose list already lives in `soft_reset`. Then:

- `ESC H` (`HTS`) sets the stop at the current column, so the tracker needs the cursor column, which the emulator has.
- `CSI g` / `CSI 0 g` clears the stop at the cursor; `CSI 3 g` clears all.
- `\t`, `CSI I` (`CHT`, forward *n* stops) and `CSI Z` (`CBT`, back *n* stops) move the cursor to the computed column, which the emulator applies as a `CSI n G` (`CHA`) on the rewritten stream — the backend dispatches `G`, so this is a rewrite rather than a new grid operation, the same way DEC Special Graphics became a glyph substitution.

Resize is the one interesting case: a grid that grows needs stops for the new columns. Take the simple answer — extend with the default every-eighth pattern past the old width, leaving existing stops alone — and write it in the rustdoc as a decision rather than leaving a reader to infer it.

Scope note for the PR: `\t` currently reaches the backend and is handled by its fixed eight, so the rewrite must take over `\t` as well or the two will disagree the moment a custom stop exists. That is the part to get right first, and a test that sets one stop and then uses a plain `\t` is the one that proves it.

**Done when** — `HTS` sets a stop that a following `\t` honours; `TBC` clears one stop and `CSI 3 g` clears all; `CHT` and `CBT` move by whole stops; `RIS` and `DECSTR` restore the default every-eighth set; a resize extends the set in a documented way; and a test in the emulation suite covers each, including a plain `\t` after a custom stop.

bernalalexis-try:
Closes #262.

Tab stops were the vt100 backend's hardcoded eight. `HTS` (`ESC H`), `TBC`
(`CSI g`, `CSI 3 g`), `CHT` (`CSI I`) and `CBT` (`CSI Z`) all reached a
dispatch table with no entry for them and vanished — though `hts`, `tbc` and
`cbt` sit in the terminfo entry termlens hands every child. An application
that laid a table out by setting its own stops and tabbing between them drew
every column in the wrong place, and did it *silently*: every character was
still present, so `contains` passed and only a column assertion or a
whole-screen snapshot could catch it.

## What changed

- **The stop set lives in `emu/seq.rs`**, beside the character sets and the
  cursor shape — one flag per column, `cols` wide, because vt100 holds no tab
  state to keep in step with.
- **The tracker classifies, the emulator resolves.** Every one of these
  operations is relative to the cursor column, which lives in the grid, so
  `SeqTracker` emits a `TabOp` and `Vt100Emulator::apply_tabs` feeds the bytes
  it still owes the parser, reads the column, and rewrites a motion as `CHA`
  (`CSI n G`) — a sequence the backend *does* dispatch. Same shape as the DEC
  Special Graphics glyph substitution, on the same staged stream.
- **Plain `HT` is rewritten too.** This is the part the issue's scope note
  calls out: vt100 answers `\t` from its own fixed eight, so leaving it there
  would have drawn a table half from each set of stops the moment an
  application set one.
- **`RIS` and `DECSTR` restore the every-eighth default**, and a resize
  extends the set into its new columns with that same pattern while leaving
  the stops it already had alone — written up as a decision in the rustdoc,
  `docs/DESIGN.md` and the README rather than left to be inferred.
- Scope, documented: `TBC 0` and `TBC 3` only (the rest of that family clears
  *line* tab stops, which this crate has no notion of), and back-tab moves to
  the nearest stop *strictly* left of the cursor, as xterm and alacritty do.

## Tests

`crates/termlens/tests/tabs.rs` — 11 end-to-end cases through a real PTY,
asserting **columns** rather than `contains`, since the failure this file
exists for leaves every character present and every column wrong: both
reproductions from the issue, a plain `\t` after an `HTS`, `TBC` of one stop
and of all of them, `CHT`/`CBT` counts, `RIS`, `DECSTR`, a real `resize`, and
a two-row table laid out with its own stops.

**9 of the 11 fail on `main`** (the two that pass are the reset cases, which
are vacuous while the stops are fixed). Plus unit coverage in `emu/seq.rs`
for the recognition and the set arithmetic, and in `emu/vt100.rs` for the
rewrite — including a tab inside a styled span, which is what would break if
only one of the two parsers were told about the `CHA`.

## Notes on the review pass

- Rebased onto `main` (the `CHANGELOG.md` conflict with #274 is resolved: the
  entry now sits under the existing `### Fixed` heading in Unreleased).
- `a_resize_extends_the_stops_and_keeps_the_ones_it_had` pinned the *row* the
  second half of its script prints on. That row depends on whether the child's
  `read` consumed our `Enter` — and so echoed a newline — or was cut short by
  the `SIGWINCH` the resize raises, which is the very race the test's own
  comment is written around. It asserted row 1 and got row 2 on every runner.
  Now it asserts the columns, which are the same either way and are what the
  resize rule is about.
- Squashed with a DCO sign-off; the original commits carried none, so the
  author of record is the one signing. Credit kept as `Co-authored-by`.

Local gates: `fmt`, `clippy` (all three feature combinations), `test`
(default, `--no-default-features`, `--features decode`), `doc`, `gates-listed`
— all green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,23 @@ listed under a **Changed** or **Removed** heading.
 
 ### Fixed
 
+- **Custom tab stops are honoured: `HTS`, `TBC`, `CHT` and `CBT` do what
+  they say.** Stops were fixed at every eighth column — the backend's
+  hardcoded value — and all four escapes reached a dispatch table with no
+  entry for them and vanished, though `hts`, `tbc` and `cbt` are all in the
+  terminfo entry termlens hands every child. An application that laid a
+  table out by setting its own stops and tabbing between them drew every
+  column in the wrong place, and did it *silently*: the characters were all
+  present, so `contains` still passed and only a column assertion or a
+  whole-screen snapshot could catch it. The stop set now lives in the
+  sequence tracker beside the character sets, and the emulator rewrites each
+  motion as a `CHA` the backend does dispatch. Plain `\t` is rewritten too,
+  or the backend's eight and a custom stop would disagree the moment one was
+  set. `RIS` and `DECSTR` restore the every-eighth default; a resize extends
+  the set into its new columns with that same pattern and leaves the stops
+  it already had alone. `CBT` is also what `Shift-Tab` sends, so an
+  application echoing one now moves. (#262)
+
 - **docs.rs labels APIs gated by the `decode` and `insta` features.** Optional
   items no longer appear to be available in every build. (#258)
 
--- a/README.md
+++ b/README.md
@@ -270,6 +270,14 @@ design. termlens's position:
   ROMs, the other national sets — is acknowledged and reads as ASCII.
   `DECSC`/`DECRC` save and restore this state with the cursor. Locking
   shifts remain G0/G1 only (`LS2`/`LS3` are not modelled).
+- **Tab stops are the application's to set.** `HTS` (`ESC H`), `TBC`
+  (`CSI g`, `CSI 3 g`), `CHT` (`CSI I`) and `CBT` (`CSI Z`) all work, and a
+  plain `\t` honours whatever stops are set rather than a fixed eight.
+  `RIS` and `DECSTR` restore the every-eighth default, and a resize extends
+  the set into its new columns with that pattern while leaving existing
+  stops alone. Back-tab moves to the nearest stop *strictly* left of the
+  cursor, as xterm does. Only `TBC 0` and `TBC 3` are modelled; the rest of
+  that family clears *line* tab stops, which this crate has no notion of.
 - `wait_frame` needs the application to bracket its repaints in DEC 2026
   synchronized updates, and only the last 8 completed frames are retained;
   everything else waits with `wait_until`, under the three rules in
--- a/crates/termlens/src/emu/seq.rs
+++ b/crates/termlens/src/emu/seq.rs
@@ -38,6 +38,19 @@
 //! emulator rewrites: the bytes the
 //! parsers see are then a translated stream rather than a sub-slice of the
 //! read, which `emu/vt100.rs` stages.
+//!
+//! The same split holds the **tab stops** (see [`TabStops`]). vt100 has a
+//! hardcoded eight and no way to be told otherwise, so `HTS`, `TBC`, `CHT`
+//! and `CBT` reached its dispatch table and vanished — and an application
+//! that lays a table out by setting its own stops, which is what the
+//! capabilities are for, drew every column in the wrong place. The set
+//! lives here; the *cursor column* every one of those operations needs
+//! lives in the grid, so the tracker emits a [`TabOp`] and the emulator
+//! resolves it against the position it holds, rewriting a motion as `CHA`
+//! (`CSI n G`) — a sequence the backend does dispatch, the same way DEC
+//! Special Graphics became a glyph substitution. Plain `HT` is rewritten
+//! too, or vt100's fixed eight and these stops would disagree the moment an
+//! application set one.
 
 use std::sync::Arc;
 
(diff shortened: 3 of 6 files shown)


vyncint (changes requested):
Welcome, and thank you for taking this one on — it is one of the harder issues in the queue, and the part you have written is genuinely good.

Before the change requests, so you know where you stand: **your work is correct.** I completed the one missing piece locally and ran it. All ten of your tests pass, the full suite is green, clippy is clean in all four feature configurations, MSRV 1.85 builds, and the docs build with warnings denied. The `TabStops` design, the `TabOp` split between tracker and emulator, the strictly-left back-tab, the resize rule, the clamping — all of it is right, and the comments explain *why* in the way this repository asks for. Your CHANGELOG entry is in `[Unreleased]`, which four other open PRs got wrong today.

So this is not "start over". It is "one file did not make it, and the way you uploaded it needs to change".

---

## 1. The blocker: `crates/termlens/src/emu/vt100.rs` is not in the PR

Your PR changes five files. The sixth one — the emulator half your `docs/DESIGN.md` paragraph describes in detail — is missing, so the crate does not compile:

```
error[E0061]: this function takes 2 arguments but 1 argument was supplied
  --> crates/termlens/src/emu/vt100.rs:72:22
   |
72 |             tracker: SeqTracker::new(capture),
   |                      ^^^^^^^^^^^^^^^--------- argument #2 of type `u16` is missing

error[E0004]: non-exhaustive patterns: `SeqEvent::Tabs(_)` not covered
  --> crates/termlens/src/emu/vt100.rs:230:30
```

That single failure is why `clippy`, `test`, `features`, `msrv` and `skill` are all red. They are not five problems; they are one. (`skill` compiles the agent guide's examples against the crate, so it fails whenever the crate does.)

If you still have your local `vt100.rs`, include it and skip to §2 — but please check the ordering note at the end of this section first, because it is the one place this is easy to get subtly wrong. If you need to rebuild it, it is three edits.

**Edit 1** — the constructor, around line 72. The tab-stop set is one flag per column, so it needs the width:

```rust
tracker: SeqTracker::new(capture, cols),
```

**Edit 2** — `set_size`, around line 413. The set has to follow a resize:

```rust
fn set_size(&mut self, rows: u16, cols: u16) {
    self.parser.screen_mut().set_size(rows, cols);
    self.shadow.set_size(rows, cols);
    // The tab-stop set is one flag per column, so it follows the width.
    self.tracker.set_cols(cols);
    // A resize can push rows into history on its own.
    self.capture_scrolled_rows();
}
```

**Edit 3** — a new arm in the `match self.tracker.step(byte)` inside `process`, next to the `SeqEvent::SoftReset` arm:

```rust
SeqEvent::Tabs(op) => {
    // Everything *before* this byte, so the column read is the one the
    // operation is relative to. It matters for plain HT alone: vt100
    // acts on it and would have moved the cursor to its own fixed
    // eight before we looked.
    self.feed_staged(&bytes[fed..i]);
    let (_, col) = self.parser.screen().cursor_position();
    let target = self.tracker.tab_op(op, col);
    self.feed_staged(&bytes[i..=i]);
    fed = i + 1;
    if let Some(target) = target {
        self.feed(format!("\x1b[{}G", target.saturating_add(1)).as_bytes());
    }
    None
}
```

**The ordering note, which is worth the paragraph.** My first attempt fed the operation's own byte *before* reading the cursor column — the obvious reading of "feed what you owe the parser first" — and seven of your ten tests failed. The reason is the case your own DESIGN paragraph singles out: vt100 ignores `HTS`, `TBC`, `CHT` and `CBT`, but it *does* act on a plain `HT`, so by the time that byte has been fed the cursor has already jumped to vt100's fixed eight and the column you read is the wrong base. Reading before feeding is correct for all five operations, because no escape *prefix* moves the cursor either. Splitting the feed at `i` rather than `i + 1` is the whole fix, and it is invisible until you run the tests.

That is also the best evidence that your tests are doing their job: they caught a real ordering bug in an implementation that compiled fine.

## 2. `cargo fmt` — one spot

`crates/termlens/src/emu/seq.rs:1151`. rustfmt wants this broken across lines:

```rust
let ps = if params_empty { 0 } else { self.csi_first_param };
```

Do not hand-edit it — run `cargo fmt --all` and it is done. That is the only formatting complaint in the whole PR.

## 3. Sign-off, commit messages, and the upload workflow

Your four commits are all titled `Add files via upload` and none carries a `Signed-off-by` line, so `commit-policy` fails. Both come from the same cause: GitHub's web upload button cannot sign off a commit and names them for you. It is also almost certainly why `vt100.rs` was left behind — five files is one too many to shepherd through that dialog.

The fix is to work from a real clone, which §4 walks through. What CI needs at the end:

- **One commit**, signed off with `git commit -s`. That appends `Signed-off-by: Your Name <your@email>` and is how you certify the [DCO](https://developercertificate.org) — see [CONTRIBUTING §5](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md#5-developer-certificate-of-origin-dco). The sign-off email must match the commit author email.
- **A [Conventional Commit](https://www.conventionalcommits.org/) subject**, imperative and under 72 characters. For this change: `fix(emu): honour the tab stops an application sets`.
- **A PR title in the same form.** This repository squash-merges, so the PR title becomes the commit subject on `main`. `Fix #262` would land there as-is.

## 4. Step by step, from where you are

Everything you have already written is recoverable — it is sitting on `main` in your fork. This pulls it onto a proper branch, adds the missing file, and updates this same PR.

```sh
# 1. A real clone (the web UI cannot sign a commit).
git clone https://github.com/bernalalexis-try/termlens
cd termlens
git remote add upstream https://github.com/vyncint/termlens
git fetch upstream

# 2. A branch off current upstream main. Working on your fork's `main`
#    is what left you with no way to keep this PR separate from the next.
git switch -c fix/262-tab-stops upstream/main

# 3. Recover the five files you already uploaded.
git checkout origin/main -- CHANGELOG.md README.md docs/DESIGN.md \
    crates/termlens/src/emu/seq.rs crates/termlens/tests/tabs.rs

# 4. Make the three edits from §1 to crates/termlens/src/emu/vt100.rs.

# 5. Format, then run what CI runs.
cargo fmt --all
cargo test --workspace --all-features
cargo clippy --workspace --all-targets --all-features -- -D warnings

# 6. One signed-off commit.
git add -A
git commit -s -m "fix(emu): honour the tab stops an application sets"

# 7. Check the two policy gates before pushing — they take a range.
.github/scripts/check-dco.sh upstream/main..HEAD
.github/scripts/check-no-ai-attribution.sh upstream/main..HEAD

# 8. Update this PR in place.
git push --force-with-lease origin fix/262-tab-stops:main
```

Step 8 pushes your branch onto your fork's `main`, which is what this PR tracks, so #273 updates rather than a new one opening. If you would rather start a clean PR, push to `fix/262-tab-stops` instead and open one from that branch; close this one and say so, and nothing is lost.

Then edit the PR title to `fix(emu): honour the tab stops an application sets`, and rewrite the description to say what the PR *does* rather than what you would do — the current text is in the conditional ("I'd add a TabStops struct…"), which reads as a plan rather than a change. `Closes #262` in the body links them so the issue closes on merge.

The full gate list is in [CONTRIBUTING §1](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md#1-dev-setup). Everything there is reproducible locally, which is the point of writing it out.

## 5. Two small things for when the rest is green

- **`docs/DESIGN.md`.** Your paragraph says the emulator resolves a `TabOp` "feeding the bytes it still owes the parser first, so the position it reads is current". After §1 you know that is precise only if "owes" excludes the operation's own byte. One clause naming plain `HT` as the reason would save the next reader the hour it cost me.
- **`tests/tabs.rs`.** Ten cases, and the one gap is the resize rule you documented in three places. I checked it by hand and your implementation is right — a stop set at column 3, then a widen from 24 to 40 columns, keeps the custom stop and gives the new columns the every-eighth pattern. It deserves a test, because it is the rule most likely to be broken by a later change and the only one nothing currently pins.

---

None of this is a reflection on the work. The tracker, the tests and the docs are the hard three-quarters and you got them right; what is left is one file and the mechanics of getting it here. Push when you are ready and I will re-review.

bernalalexis-try:
Implements custom tab stops for #262. The stop set now lives in seq.rs, and the emulator rewrites every tab movement as CHA, including plain \t. Added eleven tests plus documentation.

Following review, I added the missing vt100.rs changes and fixed formatting, docs, and a resize test.

vyncint (changes requested):
This is a big step forward, and you got the hard part right.

The missing `vt100.rs` is here, and — the part I want to call out — you found the ordering trap on your own and explained it in the doc comment on `apply_tabs` better than my review did. "Reading first is correct for all five, since no escape prefix moves the cursor either" is exactly the reasoning, and it is the sentence a future reader needs. You also added seven unit tests including the shadow-parser one, and the resize test I asked for.

I ran your branch against current `main`: `cargo fmt`, clippy with `-D warnings` in every feature configuration, the full workspace suite, and MSRV 1.85 all pass, and ten of your eleven tab tests pass. **The implementation is finished and correct.** What is left is one test line and the git mechanics.

---

## 1. One failing test, and it is not the behaviour

`a_resize_extends_the_stops_and_keeps_the_ones_it_had` fails — but look at what it caught:

```
assertion `left == right` failed: the custom stop survives:
size: 40x4  cursor: 3,4
READY
   a                            b
DONE
  left: Some((2, 3))
 right: Some((1, 3))
```

The **column is 3**, which is what the test is about: the custom stop survived the widen. The **row** is 2 where you expected 1, because the shell echoes the Enter that releases `read _`, and that echo costs a line. Your resize rule works; the assertion is just pinned to a row it does not care about.

The fix is one you already built. Line 25 of that file has `col_of`, with a comment saying a column is the only thing these tests assert — and every other test in the file uses it. This is the one place that calls `find` directly:

```rust
// crates/termlens/tests/tabs.rs, in a_resize_extends_the_stops_and_keeps_the_ones_it_had
assert_eq!(col_of(&s, "a"), Some(3), "the custom stop survives:\n{s}");
assert_eq!(
    col_of(&s, "b"),
    Some(32),
    "and column 25 tabs on to the every-eighth stop at 33:\n{s}"
);
```

I applied exactly that locally and the file goes to 11 passed, 0 failed, and the whole workspace stays green.

Worth keeping the reasoning in mind rather than just the edit: a row number in a PTY test depends on how much the shell echoed before your output, which is not what any of these tests are measuring. Your helper already encodes that judgement. This assertion just did not go through it.

## 2. The sign-off, which is now the only other blocker

`commit-policy` still fails. Your six commits are all `Add files via upload` with no `Signed-off-by` line, because GitHub's web upload button cannot add one — it writes the message for you and commits as GitHub rather than as you.

Everything else on the PR is green now, so this and §1 are the whole remaining list.

## 3. `main` has moved, so a rebase is needed

Six PRs merged today, including four that touched `[Unreleased]` in `CHANGELOG.md`. Your entry will conflict there. Nothing else in your PR conflicts — I checked by merging it onto current `main` file by file.

When you resolve it, put your entry under the existing `### Fixed` heading, after the docs.rs one. `[Unreleased]` now has **Added**, **Changed** and **Fixed** sections; keep that order.

## 4. Step by step

Your work is all in your fork, so this recovers it, adds the fix, and lands it as one signed-off commit.

```sh
# 1. A real clone. The web UI cannot sign a commit, which is the blocker.
git clone https://github.com/bernalalexis-try/termlens
cd termlens
git remote add upstream https://github.com/vyncint/termlens
git fetch upstream

# 2. A branch off current upstream main.
git switch -c fix/262-tab-stops upstream/main

# 3. Recover the six files you uploaded.
git checkout origin/main -- CHANGELOG.md README.md docs/DESIGN.md \
    crates/termlens/src/emu/seq.rs crates/termlens/src/emu/vt100.rs \
    crates/termlens/tests/tabs.rs

# 4. Reopen CHANGELOG.md. Your entry came from a main that no longer
#    exists, so move it under the current "### Fixed", after the docs.rs
#    entry, and delete any duplicated headings step 3 brought with it.

# 5. Make the one-line fix from §1 in crates/termlens/tests/tabs.rs.

# 6. Run what CI runs.
cargo fmt --all
cargo test --workspace --all-features
cargo clippy --workspace --all-targets --all-features -- -D warnings

# 7. One commit, signed off. -s is what adds the Signed-off-by line.
git add -A
git commit -s -m "fix(emu): honour the tab stops an application sets"

# 8. Check the policy gates yourself; both take a range.
.github/scripts/check-dco.sh upstream/main..HEAD
.github/scripts/check-no-ai-attribution.sh upstream/main..HEAD

# 9. Update this PR in place.
git push --force-with-lease origin fix/262-tab-stops:main
```

Step 9 pushes onto your fork's `main`, which is the branch this PR tracks, so #273 updates rather than a new one opening.

Then two edits on the PR page itself:

- **Title** → `fix(emu): honour the tab stops an application sets`. This repository squash-merges, so the PR title becomes the commit subject on `main`; `Fix #262` would land there verbatim.
- **Description** → say what the PR does rather than what you would do. It is still in the conditional ("I'd add a TabStops struct…"), which reads as a proposal. A few sentences on what is implemented, plus `Closes #262` so the issue closes on merge.

## 5. One optional thing

`docs/DESIGN.md` still says the emulator resolves a `TabOp` "feeding the bytes it still owes the parser first". After what you worked out, you know that is precise only if "owes" stops short of the operation's own byte — and your `apply_tabs` comment now says exactly why. One clause there pointing at plain `HT` would make the design note as clear as the code. Not a blocker; take it or leave it.

---

To be clear about where this stands: the emulation, the tests and the docs are done and they are good work. What is between here and merge is one assertion and a `git commit -s`. Push when you are ready and I will re-review.

vyncint (approved):

PR #276, opened 2026-09-08
test: replace the shell with the emit fixture at 72 of 99 sites

### LINKED ISSUE #249: The suite drives `/bin/sh` at 98 sites, so every test depends on the host shell and none can run where there is none
**Today** — the integration suite spawns `/bin/sh` (or `sh`) at 98 sites across 17 of its 20 test files, each with a `printf` script as the program under test — 208 `printf` invocations in all. Measured against 0.9.0:

```
$ grep -c 'spawn("/bin/sh")\|spawn("sh")' crates/termlens/tests/*.rs | awk -F: '{s+=$2} END {print s}'
98
```

Every escape-sequence test is written as `printf '\033(0lqk'`; every "print, then hold the terminal open" is `printf …; read _`. The five fixtures in `fixtures/` cover only the shapes that need a real event loop: crossterm input, resize, images, width.

**Why it is worth fixing** — three reasons, in order of weight.

The shell is a variable in every test that is not about the shell. Which `sh` runs — dash on Debian, bash in POSIX mode on macOS, busybox elsewhere — decides how `printf` treats `\033` and `%s`, what `read _` does at EOF, and how fast `while :; do …; sleep 0.01; done` iterates. The stress workflow caught exactly this in #243: a test's premise held only until the loop's `sleep` took longer to spawn than the stillness the test was measuring. The shell's process model leaked into a wait-semantics test.

It is why the suite cannot run where there is no `/bin/sh`. Windows is the obvious case — #149 names this as the first of its four steps — but so is any minimal container image.

And it is a barrier for contributors who think in Rust: `printf '\033)0\016lqk\017 lqk \016x\017'` asks its reader to decode octal escapes and shell quoting before the terminal semantics the test is about.

**Fix** — one more fixture binary, `fixtures/emit` (name open), replacing the `sh -c 'printf …; read _'` idiom with a small argument language and no shell in between:

```
emit 'ESC ( 0' lqk NL 'ESC ( B' --wait      # print, then hold stdin open until Enter
emit --raw 'caf\xe9' ' done' NL             # raw bytes, for the non-UTF-8 cases
emit first --sleep 2s ' second' --wait      # timed output, for the wait tests
```

Concretely: an `enum Step { Text, Esc, Csi, Raw, Sleep, Wait, Exit(code) }` parsed from argv, written to stdout with one `write_all` per step and a flush, and `--wait` reading one line from stdin the way every fixture already does (`docs/DESIGN.md` §2, the instant-exit caveat). Then a `tests/common` helper `emit(steps) -> Terminal`, and a mechanical migration of the `sh(...)` helpers one test file per commit, so a reviewer diffs one file's scripts against their `emit` form. Tests that are *about* a shell (`stty`, a `read` loop that echoes) keep `/bin/sh`, with a comment saying why.

Two rules for the fixture's `//!` header: no timing inside the binary except the explicit `--sleep`, and std only, so the fixture cannot become a second thing the suite tests.

**Done when** — every `printf`-only script in `crates/termlens/tests/` goes through the fixture; `grep -c 'spawn("/bin/sh")' crates/termlens/tests/*.rs` counts only tests that are about a shell, each with a comment saying so; the fixture is std-only with its steps documented in its header; and the stress workflow passes on the migrated suite.

vyncint:
Closes #249. Step 1 of #149.

## What this is

`fixtures/emit` — a std-only program that writes exactly the bytes its arguments describe, then waits, sleeps or exits as told. One argument per step; the steps are the crate doc of `fixtures/emit/src/main.rs`:

```text
TEXT  NL  CR  --text WORD        literal bytes
--esc BYTES  --csi BYTES         ESC …  /  ESC [ …
--raw SPEC                       \e \n \r \t \a \\ \xNN
--sleep DUR  --wait  --exit N    the only timing, the instant-exit guard, the status
--echo-line  --echo              read a line and write it back / copy stdin until EOF
--seq N  --cwd  --pid            the flood, the directory, the process id
--env NAME  --environ            one variable or all of them, sorted
--loop                           run what precedes once, what follows forever
```

`tests/common::spawn_emit(builder, steps)` runs it from any builder. Then the migration, **one test file per commit** so each diff is one file's scripts against their `emit` form: tabs, styles, charset, utf8, state, frames, process, builder_validation, observe, timeouts, stable, basic, concurrency, scrollback, input, queries, fixtures. Where a test used to be one `printf`, it is one `--raw`, with `\033` read as `\e` — a reviewer diffs the escape sequences directly.

## The count

`grep -c 'spawn("/bin/sh")\|spawn("sh")' crates/termlens/tests/*.rs`: **99 → 27**, and every one of the 27 has a comment or file header saying why it is a shell:

| file | left | why |
|---|---|---|
| `queries.rs` | 13 | read a reply byte for byte under `stty -icanon -echo` |
| `input.rs` | 7 | read the wire (mouse, paste, cursor keys) under `stty` |
| `drain.rs` | 3 | flood + raw mode; the deadlock the harness must never cause |
| `backpressure.rs` | 2 | `dd` an exact byte count of replies under `stty` |
| `basic.rs` | 1 | `kill -TERM $$` — a process signalling itself |
| `process.rs` | 1 | `trap … TERM` — a signal handler |

The 25 raw-mode sites share one blocker: **reading what termlens typed back needs `-icanon -echo`, and std has no `tcsetattr`.** That is the fixture's stated rule (std only, no timing but `--sleep`) doing exactly what it was written to do — the shell is the honest tool for a test that is about the line discipline. If those 25 should go too, the step is `--raw-mode` on the fixture via `libc` (one `tcsetattr`) or crossterm, and that is a decision about the rule, not about the migration. Flagging rather than deciding.

## Two tests changed shape on purpose

- `basic.rs` `send_str_and_enter_round_trip…` waited for `got: hello`, which the line discipline's echo of the typed `hello` could satisfy on its own. It now waits for `hello back` — the fixture's read plus a suffix the echo cannot produce.
- `builder_validation.rs` `a_bare_program_name_under_env_clear…` used `sh` as the bare name and `/usr/bin:/bin` as the PATH remedy. It now uses the fixture as the bare name and the fixture's own directory as the remedy: the same refusal, no Unix path baked in.

Everything else asserts what it asserted before.

## Verified

Every gate in CONTRIBUTING §1 locally, including the cross-target clippy; full suite three feature configurations. The `windows` and `stress` workflows are dispatched against this branch — results in a comment below, since #249's done-when names the stress workflow and #149's step 4 is what the Windows count is for.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -129,6 +129,10 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "75b325c5dbd37f80359721ad39aca5a29fb04c89279657cffdda8736d0c0b9d2"
 
 [[package]]
+name = "emit"
+version = "0.9.0"
+
+[[package]]
 name = "encode_unicode"
 version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
--- a/crates/termlens/tests/backpressure.rs
+++ b/crates/termlens/tests/backpressure.rs
@@ -4,6 +4,10 @@
 //! The other half of this story — what a harness can know about an
 //! application that never reads its replies — lives in `drain.rs`, next to
 //! the deadlock it must never cause.
+//!
+//! Stays on `/bin/sh` on purpose (#249): reading replies byte for byte
+//! needs the terminal in raw mode, which the std-only `emit` fixture cannot
+//! set — see the note in `drain.rs`.
 
 use std::time::Duration;
 
--- a/crates/termlens/tests/basic.rs
+++ b/crates/termlens/tests/basic.rs
@@ -1,10 +1,10 @@
-//! Integration tests against `/bin/sh`: spawning, waiting, environment
-//! control, exit codes, and the failure modes (timeout / EOF).
+//! Integration tests against the `emit` fixture: spawning, waiting,
+//! environment control, exit codes, and the failure modes (timeout / EOF).
 //!
 //! These run headless — CI runners have no TTY, the harness makes its own.
 //!
-//! Pattern note: every script that must *print something we assert on*
-//! ends with a `read` guard, and we send Enter only after the assertion.
+//! Pattern note: every program that must *print something we assert on*
+//! ends with a `--wait`, and we send Enter only after the assertion.
 //! Output written immediately before exit can be discarded by macOS's PTY
 //! teardown (docs/DESIGN.md §2); keeping the child alive until the harness
 //! has seen the bytes makes these tests deterministic on every platform.
@@ -16,15 +16,17 @@ use std::{
 
 use termlens::{Error, Key, Terminal};
 
-const SH: &str = "/bin/sh";
+mod common;
 
-fn sh(script: &str) -> termlens::Result<Terminal> {
-    Terminal::builder().args(["-c", script]).spawn(SH)
+/// The `emit` fixture from the default builder; steps are documented in
+/// `fixtures/emit/src/main.rs`.
+fn emit(steps: &[&str]) -> termlens::Result<Terminal> {
+    common::emit(steps)
 }
 
 #[test]
 fn echo_reaches_the_screen_and_child_exits_cleanly() -> termlens::Result<()> {
-    let mut t = sh("echo hello from a real PTY; read guard")?;
+    let mut t = emit(&["hello from a real PTY\n", "--wait"])?;
     t.wait_until(|s| s.contains("hello from a real PTY"))?;
     t.send(Key::Enter)?;
     let status = t.wait_exit()?;
@@ -35,9 +37,9 @@ fn echo_reaches_the_screen_and_child_exits_cleanly() -> termlens::Result<()> {
 
 #[test]
 fn exit_codes_are_reported() -> termlens::Result<()> {
-    // The `read` keeps the exit from racing PTY setup; the line discipline
-    // buffers our Enter even if it lands before `read` starts.
-    let mut t = sh("read guard; exit 7")?;
+    // The `--wait` keeps the exit from racing PTY setup; the line discipline
+    // buffers our Enter even if it lands before the read starts.
+    let mut t = emit(&["--wait", "--exit", "7"])?;
     t.send(Key::Enter)?;
     let status = t.wait_exit()?;
     assert!(!status.success());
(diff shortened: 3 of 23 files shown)


vyncint:
## `windows.yml` against this branch — [run 34179522277](https://github.com/vyncint/termlens/actions/runs/34179522277)

**345 passed, 92 failed** — against 274 / 163 on `main` after #275. Of the 92, **30 name a shell** (the 27 sites kept on purpose, plus the two `wait_frame` doctests and the `tiocgwinsz` python one); the other 60 are tests that used to die on `/bin/sh` before they could say anything and now run far enough to hit ConPTY. They line up with the step-3 table on #149 — every `frames.rs`, `graphics.rs`, and `queries.rs` reply test; mouse modes, focus, OSC 8 ids, DECSTR, DEC Special Graphics in `charset.rs`, the raw bytes in `utf8.rs` — with **one new finding** worth adding to that table:

- **Child exit is not EOF on ConPTY.** `a_departed_child_is_refused_identically_on_every_platform`, `typed_input_to_a_departed_child_is_a_typed_error`, `a_mouse_click_at_a_departed_child_blames_the_child`, `resize_after_the_child_exits_is_refused_like_send` and `waits_fail_fast_on_eof…` all fail the same way: the child has exited, and the harness has not seen EOF, because the output pipe stays open until the pseudoconsole is closed. On Unix the liveness check reads EOF off the master; on Windows it would have to read the process handle. That is a harness change for step 4, not a test change.

Nothing here is a regression on Unix — the suite, all three feature configurations, is green on the PR — and nothing here is the fixture misbehaving on Windows either: `emit` built and ran there, `--wait` blocked on cooked-mode input as `read` did, `--environ` reported the environment, and the two `bin!`-style absolute-path spawns work. The `stress` run is dispatched too; numbers when it lands.

vyncint:
## `stress.yml` against this branch — [run 34179524611](https://github.com/vyncint/termlens/actions/runs/34179524611)

All ten shards green: ubuntu and macOS at 1, 2, 4, 8 and 16 threads, the suite repeated per shard as the workflow weights them. That is #249's last done-when — *the stress workflow passes on the migrated suite* — and it is the one that mattered most, since the shell's process model was what leaked into the wait-semantics tests before (#243).

Issue #278, opened 2026-09-08
inspect.rs rebuilds the example from every test, so two tests can race a relink and one finds no binary


vyncint:
**Today** — `tests/inspect.rs` gets the example binary through `inspect_bin()`, which runs `cargo build --example inspect` and returns `target/<profile>/examples/inspect`. Every test in the file calls it, and the tests run in parallel, so two `cargo build`s race: while one is relinking the example — unlinking and rewriting the file — the other's `Command::new(bin).output()` finds nothing there. Seen once on the macOS leg of #277's first push, in a file that PR did not touch:

```
---- inspect_prints_its_usage_for_help_and_for_a_missing_program stdout ----
thread '…' panicked at crates/termlens/tests/inspect.rs:28:10:
failed to run the inspect example: Os { code: 2, kind: NotFound, message: "No such file or directory" }
test result: FAILED. 5 passed; 1 failed
```

Green on the re-run, which is the shape of a race and not of a bug in the example.

**Why it is worth fixing** — a flaky required check teaches people to re-run CI, which is the one habit `docs/DESIGN.md` §2 is written against. And the fix already exists in the same directory: `tests/common::fixture_bin` serializes its `cargo build -p` behind a `BUILT: Mutex<Vec<String>>` so each fixture is built once per test process, for exactly this reason (its comment: *"`cargo test` links test harnesses into `deps/` but does not refresh the plain bin artifact"*). `inspect_bin()` predates that helper and never picked it up.

**Fix** — the same `OnceLock`/mutex shape around the `cargo build --example inspect`, so the example is built once per test process and every test after the first only reads the path. Or teach `fixture_bin` to take an `--example`, and delete the second copy of the idea.

**Done when** — `inspect_bin()` cannot run two builds in one test process, and the stress workflow's macOS shards are green across a full run with the inspect tests in them.

PR #279, opened 2026-09-08
ci: run the suite on windows-latest, and say what Windows is

vyncint:
Refs #149 — everything that remained of step 4, plus the last of #249 and #278.

## What lands

**The fixture takes the last 27 shell sites.** `emit` gains the steps the reply readers needed — `--raw-mode`/`--no-icrnl` (one `tcsetattr`), `--read`/`--skip`/`--read-hex`/`--read-quiet`/`--read-count`, `--winsize`, `--kill-self`, `--on-term … --idle` — via `libc`, Unix only, and its header rule is rewritten to say exactly that: *std, plus libc for the terminal-mode, ioctl and signal steps and nothing else.* `grep -c 'spawn("/bin/sh")\|spawn("sh")' crates/termlens/tests/*.rs` is **zero**. Off Unix the steps are accepted and do nothing; the tests that depend on them are Unix-only for other reasons (below).

**79 tests are `#[cfg_attr(windows, ignore = "…")]`**, each with the row of the probe table it fails on:

| reason | tests |
|---|---|
| ConPTY closes a DEC 2026 bracket before the content it wrapped | 26 (`frames.rs`, form-echo-driven `input.rs`, one each in `observe.rs`/`timeouts.rs`) |
| ConPTY answers or eats the child's queries itself | 20 (`queries.rs`, `drain.rs`, `backpressure.rs`) |
| ConPTY does not forward kitty or sixel payloads | 12 (`graphics.rs`, `observe.rs`) |
| the wire is read in raw mode, a `tcsetattr` | 6 (`input.rs`) |
| ConPTY does not forward mode sets / DECSTR / turns 1004 on for itself / chooses OSC 8 ids | 3 + 2 + 1 + 1 (`state.rs`, `tabs.rs`, `input.rs`) |
| ConPTY translates charset designations itself | 3 (`charset.rs`) |
| Rust's console stdio refuses non-UTF-8 | 3 (`utf8.rs`, `fixtures.rs`) |
| Windows has no signals / TIOCGWINSZ is a Unix ioctl | 1 + 1 |

Two tests were wrong rather than impossible: they typed a bare LF to end a line, and ConPTY's cooked input ends one at CR. They send `Enter` now and pass on Windows. The two `wait_frame` doctests are Unix-only like the `Signal` one.

**The leg**: `windows (non-required)` in `ci.yml`, `continue-on-error: true`, outside `required-green`. Promoting it is two edits, written next to it. `windows.yml` stays as the dispatchable version with the probe and the report.

**The sentence**: README's "Unix only for now" becomes *Windows: screen assertions yes, frame assertions no*, with the list of what ConPTY renders away. DESIGN §1 gets the mechanics — the preamble, the `6n` handshake the child waits on, exit ≠ EOF. CHANGELOG gets the user-facing entry.

**#278**: `inspect.rs` builds the example once per test process (`OnceLock`), the guard `fixture_bin` already had.

## Verification

Unix: full suite, three feature configurations, both clippy targets, MSRV, docs, deny, gates-listed, zizmor. Windows: `windows.yml` dispatched against this branch after the ignore pass — the expected outcome is **0 failed**; result in a comment below, along with the new `windows (non-required)` job's own run on this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,6 +111,27 @@ jobs:
           targets: x86_64-pc-windows-msvc
       - run: cargo clippy --workspace --all-targets --all-features --target x86_64-pc-windows-msvc -- -D warnings
 
+  # The Windows leg (#149): the whole suite over ConPTY, with the tests that
+  # ConPTY cannot honour marked `#[cfg_attr(windows, ignore = "…")]`, each
+  # with its reason. Non-required while it earns it — `continue-on-error`
+  # keeps a ConPTY surprise from blocking a Unix fix, and required-green does
+  # not list it. Promoting it is two edits: drop continue-on-error, add it
+  # to required-green's needs. windows.yml is the dispatchable version with
+  # the ConPTY probe and the report.
+  windows:
+    name: windows (non-required)
+    runs-on: windows-latest
+    continue-on-error: true
+    timeout-minutes: 45
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          persist-credentials: false
+      - uses: dtolnay/rust-toolchain@6c977a6ca4077a0ceb28ffbe03f59d46e9ac8772 # v1
+        with:
+          toolchain: stable
+      - run: cargo test --workspace --all-features
+
   # CONTRIBUTING §1 promises that every gate above is reproducible locally
   # and that the workflow is the source of truth when the two disagree. It
   # disagreed twice before anyone noticed (#227); this is the check that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,17 @@ listed under a **Changed** or **Removed** heading.
 
 ### Added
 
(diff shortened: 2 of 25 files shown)


vyncint:
## Results

- **`windows.yml` against this branch** — [run 34185464636](https://github.com/vyncint/termlens/actions/runs/34185464636): **358 passed, 0 failed, 82 ignored.** Every ignore carries its reason.
- **`windows (non-required)` on this PR**: green. Its first run failed one test the dispatch workflow had passed — `env_clear_blocks_inheritance…` asserted `HOME` in the parent environment, which Windows has only when a shell put one there (the dispatch step runs under bash; the CI leg under pwsh). It probes `USERPROFILE` on Windows now.
- **`stress.yml` against this branch** — [run 34185491584](https://github.com/vyncint/termlens/actions/runs/34185491584): all ten shards green, ubuntu and macOS at 1/2/4/8/16 threads, with the fixture's new raw-mode and read steps under every reply reader.
- Unix legs, both clippy targets, MSRV, docs, deny, gates-listed, zizmor: green.

PR #291, opened 2026-09-08
ci: hunt flakes on Windows too, where the exit edge is synthesized

### LINKED ISSUE #290: The stress workflow skips the one platform whose liveness is synthesized rather than observed
**Today** — `stress.yml` runs the suite in a 100-iteration loop over `ubuntu-latest` and `macos-latest`, sharded across five `--test-threads` values. Measured against 0.10.0: `windows-latest` appears nowhere in it, while `ci.yml`'s `windows` job became a **required** check in #288/#280. The README calls the stress workflow "on Linux and macOS" and CONTRIBUTING §3 says the same, so the gap is documented rather than hidden — but it is still a gap on the newest and least-exercised leg.

**Why it is worth fixing** — a required check has no flake budget, and the two things holding the Windows leg up are timing rather than logic:

1. **The terminal-closed edge is manufactured there.** On Unix it is a kernel fact: EOF on the master means every slave descriptor is closed. On Windows the pseudoconsole's output pipe outlives the child, so `EXIT_CLOSES_THE_TERMINAL` + `ExitWatch` synthesize the edge — `try_wait()`, then `DRAIN_GRACE` (500 ms) of drain, then "closed". Polling plus a grace window is precisely the shape a one-in-N flake lives in, and this code has never run more than once per commit.
2. **`common::fixture_bin` shells out to `cargo build -p` from inside the tests**, on the one filesystem that refuses to relink a running `.exe`. #149's step-4 note called this out as untested at depth.

Both are exactly what the stress workflow was built to find; it found a genuine macOS PTY-device race the first day it ran, and the instant-exit teardown loss at roughly 1 in 80. A Windows flake found this way arrives with a thread count and an iteration attached. The same flake found later arrives as a red required check on someone else's unrelated pull request.

**Fix** — add `windows-latest` to the stress matrix. Three things it is not:

- not a blind matrix add: `windows-latest` defaults to `pwsh` and the iteration loop is POSIX shell, so the job needs `defaults.run.shell: bash` (Git Bash ships `seq` and arithmetic expansion, so one script still serves all three);
- not the same timeout: a Windows iteration is roughly two to three times a Linux one, so the serial shard needs headroom the Unix legs do not;
- not the same faults at the ends of the thread axis, and the comment should say so — at 1 thread Windows is hunting the exit synthesis (nothing contends, so a child can be reaped before its last write is drained), at 16 it is hunting conhost churn on a four-core box.

The 94 `#[cfg_attr(windows, ignore)]` tests stay ignored; they are what ConPTY cannot honour, not what is untested.

**Done when** — `stress.yml` runs the loop on all three OSes, a real dispatch of it against `windows-latest` has completed green, and the README's and CONTRIBUTING §3's "Linux and macOS" both name the third.

vyncint:
The stress workflow ran on Linux and macOS only, which left the leg that became **required** in 0.10 as the least exercised — and it is the one whose liveness the harness *manufactures* rather than reads.

On Unix "the terminal closed" is a kernel fact: EOF on the master means every slave descriptor is closed. On Windows the pseudoconsole's output pipe outlives the child, so `EXIT_CLOSES_THE_TERMINAL` + `ExitWatch` synthesize the edge — reap the child, drain for `DRAIN_GRACE` (500 ms), call it closed. Polling plus a grace window is the shape a one-in-N flake lives in, and nothing had run it more than once per commit. The other Windows-only path is `common::fixture_bin`, which shells out to `cargo build -p` from inside the tests, on the one filesystem that refuses to relink a running `.exe`.

**Three things this is not:**

- **Not a blind matrix add.** `windows-latest` defaults to `pwsh` and the iteration loop is POSIX shell, so the job pins `defaults.run.shell: bash`. Git Bash ships `seq` and arithmetic expansion, so one script still serves all three; on Unix the only change is an unused `pipefail`.
- **Not the same clock.** A Windows iteration costs two to three times a Linux one and the serial shard is the long pole, so Windows gets 60 minutes against 30 — a timeout that fires on a healthy run is a false alarm that costs more than the minutes it saves.
- **Not the same faults at the ends of the thread axis**, and the comment now says so: at 1 thread Windows hunts the exit synthesis rather than the macOS teardown race, at 16 it hunts conhost churn on a four-core box.

The 94 `#[cfg_attr(windows, ignore)]` tests stay ignored — that is what ConPTY cannot honour, not what is untested. README and CONTRIBUTING §3 now name the third OS.

**Proof, not just intent:** a real 20-iteration dispatch of the new workflow ran against this branch before merge — https://github.com/vyncint/termlens/actions/runs/34200079944

Closes #290

### DIFF
--- a/.github/workflows/stress.yml
+++ b/.github/workflows/stress.yml
@@ -1,7 +1,19 @@
-# Flake hunter: the whole test suite, 100 times, on both OSes.
+# Flake hunter: the whole test suite, 100 times, on all three OSes.
 #
 # A flaky test harness is worse than none — this workflow is the credibility
 # gate. Run it before every release and whenever wait/timing code changes.
+#
+# Windows joined in 0.10 (#290), when its ci.yml leg became a required check
+# (#280) and therefore acquired a flake budget of zero. It is here for one
+# reason the other two legs do not have: on Unix "the terminal closed" is a
+# kernel fact — EOF on the master means every slave descriptor is closed —
+# while on Windows the pseudoconsole's output pipe outlives the child, so
+# the harness *manufactures* the edge instead (`EXIT_CLOSES_THE_TERMINAL`,
+# `ExitWatch`: reap the child, drain for `DRAIN_GRACE`, call it closed).
+# Polling plus a grace window is the shape a one-in-N flake lives in, and
+# until now nothing ran it more than once per commit. The second reason is
+# `common::fixture_bin`, which shells out to `cargo build -p` from inside
+# the tests, on the one filesystem that refuses to relink a running `.exe`.
 #
 # The 100 are split across SHARDS machines rather than run end to end on one.
 # Three things follow, and only the last one is a cost:
@@ -36,6 +48,13 @@
 #
 # The thread count is in the job name and in the failure, so a flake arrives
 # already half-diagnosed: at 1 it is a speed race, at 16 a contention one.
+#
+# The axis transfers across platforms; the faults at its ends do not. At 1
+# thread Windows is hunting the exit synthesis above rather than the macOS
+# teardown race — nothing contends, so a child can be reaped before its last
+# write has been drained. At 16 it is hunting conhost churn: windows-latest
+# is a four-core box, so sixteen pseudoconsoles opening and closing at once
+# is genuine oversubscription rather than a figure of speech.
 name: stress
 
 on:
@@ -66,7 +85,7 @@ jobs:
       # on is most of the diagnosis, and a cancelled shard reports nothing.
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest, windows-latest]
         threads: [1, 2, 4, 8, 16]
         # Balanced by wall time rather than by count. Measured on this suite:
         # a serial iteration costs 43s against 16s at four threads and 14s at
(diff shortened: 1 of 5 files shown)


vyncint:
**Dispatch results — the leg works, and it immediately earned its keep.**

Two real dispatches of the new workflow against this branch:

| run | iterations | Windows shards | outcome |
|---|---|---|---|
| [34200079944](https://github.com/vyncint/termlens/actions/runs/34200079944) | 20 | 5/5 green | found a `queries.rs` flake on macOS @4 threads |
| [34201210388](https://github.com/vyncint/termlens/actions/runs/34201210388) | 40 | 5/5 green | found a `frames.rs` flake on macOS @1 thread |

Neither flake is Windows and neither is caused by this change — both are pre-existing macOS timing faults that a deeper loop surfaced, which is the entire point of the workflow.

- **Fixed here** (second commit): `a_query_the_app_moved_past_is_context_not_a_cause` spent one 400 ms budget on both a wait that must succeed and a wait that must expire, with a deliberate 200 ms fixture sleep in between. Root-caused, fixed by moving the short deadline onto the wait that needs it, and hammered 40× locally at 8 threads with zero failures.
- **Filed, not guessed at**: #292 — a resize notification going missing on macOS so `form-echo` never acknowledges the SIGWINCH (`last:` empty, "1 complete frame in total" after 10 s). That is not the frame-cursor gap this test was originally written for, it did not reproduce in 30 local Linux runs, and widening a deadline would suppress it rather than fix it. It deserves its own investigation and does not block this change.

**Sizing, now measured rather than estimated:** 40 iterations cost ~10 min per Windows shard, so a full 100 extrapolates to ~25 min plus the ~6 min prebuild — comfortably inside the 60 minutes this PR gives Windows, with room for a slow runner day.

PR #321, opened 2026-09-08
fix(cli): handle --version in each subcommand

### LINKED ISSUE #310: termlens <subcommand> --version is an unknown option
**Today** — `termlens --version` prints the version, but the flag is handled only in the top-level dispatch. Measured on 0.10.1 with the published binary:

```console
$ termlens --version
termlens 0.10.1
$ termlens inspect --version
termlens: unknown option "--version" (try `termlens inspect --help`)
$ echo $?
2
```

`grep -n '"--version"' crates/termlens-cli/src/main.rs` finds a single site, in `main`.

**Why it is worth fixing** — `--version` after a subcommand is what people type, and every subcommand already accepts `-h`/`--help` in that position. A small, self-contained inconsistency.

**Fix** — in `crates/termlens-cli/src/main.rs`, handle `--version` in each subcommand's flag loop exactly as `-h`/`--help` is handled, printing the same string `main` prints. Keep that string in one place rather than repeating the `format!`.

**Done when**
- [ ] `termlens inspect|diff|render --version` each print the same line as `termlens --version` and exit 0.
- [ ] `crates/termlens-cli/tests/cli.rs` covers at least one subcommand.

---

jabrailkhalil:
Fixes #310.

## Problem

`termlens --version` worked, but `termlens inspect|diff|render --version` returned exit code 2 with `unknown option`.

## Fix

- Extracted `version() -> String` helper so the formatted version string lives in one place.
- Added `--version` handling to the flag loop of each subcommand (`diff`, `render`, `inspect`), mirroring how `-h`/`--help` is handled.
- Added `subcommand_version_prints_same_string_as_top_level` test in `crates/termlens-cli/tests/cli.rs` that checks all three subcommands exit 0 and print the same string as `termlens --version`.

Signed-off-by: Jabrail <78273416+jabrailkhalil@users.noreply.github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,13 @@ listed under a **Changed** or **Removed** heading.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`termlens <subcommand> --version` is no longer an unknown option.**
+  The flag was handled only in the top-level dispatch, so `termlens inspect
+  --version` exited 2 while `-h`/`--help` worked in that position. All three
+  subcommands now print the same version string as the top-level command. (#310)
+
 ## [0.10.1] - 2026-09-08
 
 ### Changed
--- a/crates/termlens-cli/src/main.rs
+++ b/crates/termlens-cli/src/main.rs
@@ -92,6 +92,11 @@ fn fail(message: &str) -> ExitCode {
     ExitCode::from(2)
 }
 
+/// The one-line string every `--version` flag prints.
+fn version() -> String {
+    format!("termlens {}\n", env!("CARGO_PKG_VERSION"))
+}
+
 fn main() -> ExitCode {
     let mut args = std::env::args().skip(1);
     let Some(command) = args.next() else {
@@ -101,7 +106,7 @@ fn main() -> ExitCode {
     let rest: Vec<String> = args.collect();
     match command.as_str() {
         "-h" | "--help" => print(&format!("{USAGE}\n")),
-        "--version" => print(&format!("termlens {}\n", env!("CARGO_PKG_VERSION"))),
+        "--version" => print(&version()),
         "inspect" => inspect(rest),
         "diff" => diff(&rest),
         "render" => render(&rest),
@@ -164,6 +169,7 @@ fn diff(args: &[String]) -> ExitCode {
     while let Some(arg) = args.next() {
         match arg.as_str() {
             "-h" | "--help" => return print(&format!("{DIFF_USAGE}\n")),
+            "--version" => return print(&version()),
             "--color" => {
                 color = match args.next().map(String::as_str) {
                     Some("auto") => ColorWhen::Auto,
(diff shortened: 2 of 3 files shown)


vyncint (changes requested):
Thank you for this — it's a well-shaped first contribution, and it does exactly what #310 asked for rather than the minimum that would close it. Extracting `version()` instead of copying the `format!` into three more places was the right instinct, and covering all three subcommands when the issue only asked for one is the kind of thing that makes a patch easy to review.

I verified it rather than eyeballing it, so you know what's been checked:

| check | result |
|---|---|
| `--version` on all four commands | all print `termlens 0.10.1`, all exit 0 |
| `termlens inspect /bin/echo --version` | `--version` still reaches **the program**, not termlens — the non-obvious risk in this change, and it's handled because your arm sits inside the pre-program flag loop |
| `render`/`diff` normal operation | unaffected |
| DCO sign-off, no AI trailers, Conventional Commit | all clean |
| `fmt`, `clippy --all-targets --all-features`, `cargo test -p termlens-cli` | clean locally |
| CI | all 15 checks green, Windows and macOS included |
| new test × 40 runs at 4 threads | 0 failures — no instant-exit flake |

## One thing before merge

**`CHANGELOG.md` needs an entry under `[Unreleased]`.** [CONTRIBUTING §7](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md) asks for one on any user-facing change, and it's the sixth item on the PR checklist. This isn't bookkeeping: `.github/scripts/extract-changelog.sh` builds the GitHub Release notes straight from that file, so as it stands your fix would ship in 0.10.2 without appearing in the notes.

`[Unreleased]` is empty right now, so the whole edit is:

```markdown
## [Unreleased]

### Fixed

- **`termlens <subcommand> --version` is no longer an unknown option.** The
  flag was handled only in the top-level dispatch, so `termlens inspect
  --version` exited 2 with `unknown option` while `-h`/`--help` worked in
  that position. All three subcommands now print the same line the top level
  does, from one `version()` helper. (#310)
```

That's the only blocker — one commit and this is ready. Everything else above is verified and green.

## Non-blocking, entirely your call

The subcommand usage strings (`DIFF_USAGE`, `RENDER_USAGE`, `INSPECT_USAGE`) don't list `--version`, and the top-level `USAGE` gives `--help` the note *"(also after a command)"* while `--version` has no equivalent. Now that the two behave the same, making the docs say so would be consistent. Happy to merge without it if you'd rather keep the diff tight — or take it in this PR, whichever you prefer.

Thanks again for picking this one up.

vyncint (approved):

vyncint:
@jabrailkhalil
Error: Commit 56beb296c4b933ece390759f545791ab8fe1f981 has no Signed-off-by matching its author <78273416+jabrailkhalil@users.noreply.github.com>.
  subject: docs: add changelog entry for subcommand version
  fix: 'git commit --amend -s' for the last commit, or 'git rebase --signoff' for a branch, then force-push.
check-dco: inspected 2 commit(s) in dc9a073548f9669d49138afdec32e6130b5a2a28..56beb296c4b933ece390759f545791ab8fe1f981
Error: DCO check failed. Every commit must be signed off (git commit -s). See CONTRIBUTING.md §5 and https://developercertificate.org/
Error: Process completed with exit code 1.

jabrailkhalil:
Added the requested [Unreleased] / Fixed changelog entry in commit 56beb29. No local Cargo toolchain was available here; the existing CI matrix was green before this documentation-only change.

jabrailkhalil:
Follow-up: amended the changelog commit with the required DCO sign-off; the branch now points to 674227e.

PR #351, opened 2026-09-13
docs : add contibutor getting started guide

DevanshGang:
## What & why

Added a short "Where to start " section to CONTRIBUTING.md to make it easier for new contributors to get started.

this section points to good first issues , explains that early draft PRs are welcome and highlights the DCO sign-off and snapshot review requirements.

The README Contributing section now links directly to the new section.

Closes #314 
## Checklist

- [X] Linked an issue (or explained above why none exists)
- [ ] Tests added/updated for the change
- [ ] `cargo fmt --all` and `cargo clippy --workspace --all-targets --all-features` are clean
- [X] **All commits are signed off** (`git commit -s`) — see [CONTRIBUTING.md §5](../CONTRIBUTING.md)
- [X] **No AI attribution trailers** (no AI co-authors, "Generated with" footers, or bot identities) — see [CONTRIBUTING.md §6](../CONTRIBUTING.md)
- [ ] `CHANGELOG.md` updated under `[Unreleased]` (user-facing changes only)
- [ ] Snapshot changes (if any) were reviewed with `cargo insta review`, not blind-accepted

PR #353, opened 2026-09-13
fix(ci): check CONTRIBUTING links, not just the README's

### LINKED ISSUE #352: check-readme-links only reads README.md, so a broken link in CONTRIBUTING.md passes every gate
**Today** — `.github/scripts/check-readme-links.sh` is the only link gate, and
`ci.yml` runs it as `.github/scripts/check-readme-links.sh` with no argument,
so it reads `README.md` and nothing else. Measured on #351: a link to
`https://github.com/vyncint/temlens/labels/good%20first%20issue` (missing the
`r`) landed in `CONTRIBUTING.md` and all sixteen checks reported success. The
URL 404s; `curl -o /dev/null -w '%{http_code}'` returns `404` for it and `200`
for the correct spelling.

Two separate reasons it was missed, and both still stand:

1. The file is out of range. Only `README.md` is read.
2. Even pointed at the file, the second half would not have caught it. That
   half resolves paths under `https://github.com/vyncint/termlens/blob/main/`
   against the working tree; a `…/labels/…` URL, and anything on a different
   host or a misspelled one, names no path and is simply never examined.

**Why it is worth fixing** — the broken link was the one thing that section
existed to provide, and CONTRIBUTING is the file a first-time contributor
reads. It is also the file most likely to accumulate link rot, since it points
at scripts and workflows that get renamed.

**Fix** — the two halves of the script want separating, because they do not
generalise together:

- The *absolute-only* rule is README-specific on purpose. It exists because
  `crates/termlens/Cargo.toml` sets `readme = "../../README.md"` and crates.io
  rewrites relative links against the crate directory. `CONTRIBUTING.md` is not
  packaged and legitimately uses relative links today (`AGENTS.md`,
  `docs/RELEASING.md`, `docs/STABILITY.md`, `.github/scripts/check-dco.sh`,
  `.github/scripts/check-no-ai-attribution.sh`, `.github/workflows/ci.yml`).
  Pointing the current script at CONTRIBUTING would fail on all six, which is
  why #351 did not just do it.
- The *target exists* rule generalises fine, and should run over CONTRIBUTING
  too, resolving relative links against the repository root and absolute
  `blob/main/` links as it does now.

That leaves the actual hole: a URL naming no in-repo path, which today is
checked by nobody. Options, cheapest first:

- Verify that any `github.com/vyncint/<repo>` URL names a repository that
  exists in the org, offline, from a small allow-list. Catches `temlens`
  exactly, needs no network, stays deterministic.
- A real link check over external URLs. Catches more, but it is online and
  flaky, and the existing script's header argues deliberately against making
  the gate depend on GitHub answering. If it lands, it belongs in a scheduled
  workflow, not in the PR gate.

The first is probably enough. Whichever is chosen, break it on purpose first
and show it failing, per CONTRIBUTING.

**Done when**
- [ ] A misspelled `github.com/vyncint/temlens` link in `CONTRIBUTING.md`
      fails CI, demonstrated by a deliberate failing run.
- [ ] A renamed or deleted in-repo target linked from `CONTRIBUTING.md` fails.
- [ ] The README's absolute-only rule still applies to `README.md` alone, and
      CONTRIBUTING's existing relative links still pass.
- [ ] The reason the rules differ is written down where the next reader finds
      it, rather than inferred from the script.

---

*A good first contribution: shell only, no Rust. Start with
— §1 lists every gate you can run locally. Commits are Conventional Commits
and need `git commit -s` (DCO). Happy to review a draft PR early.*

jarik2014:
**What was broken**

`.github/scripts/check-readme-links.sh` read `README.md` and nothing else, so
`CONTRIBUTING.md` — the file a first-time contributor reads, and the one most
likely to accumulate link rot because it points at scripts and workflows that
get renamed — had no link check at all. `#351` proved it: a link one letter
short of the repository name sat in `CONTRIBUTING.md` and all sixteen checks
reported success.

**What this changes**

The script now takes a file list and gives each file the rule that applies:

* **an in-repo target must exist** — for every file read. Relative links are
  resolved against the repository root, which is where both files live.
* **no relative targets** — still `README.md` only. That rule is crates.io's:
  `crates/termlens/Cargo.toml` sets `readme = "../../README.md"`, so a relative
  link is rewritten against the crate directory. `CONTRIBUTING.md` is not
  packaged and legitimately links `AGENTS.md`, `docs/RELEASING.md` and
  `.github/workflows/ci.yml` relatively — pointing the old script at it would
  have failed on all of those, which is why `#351` did not simply do it.
* **a URL naming no in-repo path** — `github.com/vyncint/<repo>` is asserted
  against the four repositories in the org, offline. That is what catches
  `temlens`. It stays offline on purpose: a link checker that asks GitHub is
  online, flaky, and can report a 200 for the wrong reason, which is the
  failure mode this gate exists to avoid.

Why the two rules differ is now written down in the script header, next to the
reason for each, and the two commands are listed in `CONTRIBUTING.md` §1 with
the rest of the gates.

`tools/link-gate-selftest/run.sh` proves the gate can fail before it is trusted
to pass, in the shape `tools/semver-gate-selftest/` already uses: five fixtures
that must go red, and two runs that must stay green — the fixture that models
what `CONTRIBUTING.md` legitimately does, and this repository's own two files.
CI runs it in `gates-listed`, which is shell-only and single-OS.

**Evidence**

`Done when` item by item, from this branch (`tools/link-gate-selftest/run.sh`):

```
  ok    CONTRIBUTING.md: relative links, org link, fragment, external exit 0
  ok    CONTRIBUTING.md: a deleted in-repo target fails            exit 1
  ok    CONTRIBUTING.md: a misspelled org repository fails (#351)  exit 1
  ok    README.md: a relative target fails (crates.io rewrites it) exit 1
  ok    README.md: a missing absolute target fails                 exit 1
  ok    this repository's own two files pass                       exit 0
link gate selftest: 6 expectation(s), the gate fails when it should
```

And the same gate against the **real** files with each fault introduced by hand
and reverted, which is the deliberate failing run the issue asks for:

```
$ sed -i 's|/vyncint/termlens/labels|/vyncint/temlens/labels|' CONTRIBUTING.md
$ .github/scripts/check-readme-links.sh README.md CONTRIBUTING.md
README.md: 24 distinct link target(s)
CONTRIBUTING.md: 15 distinct link target(s)
CONTRIBUTING.md LINK: github.com/vyncint/temlens names no repository in the org
$ echo $?
1

$ sed -i 's|](docs/RELEASING.md)|](docs/RELEASING-old.md)|' CONTRIBUTING.md
$ .github/scripts/check-readme-links.sh README.md CONTRIBUTING.md
CONTRIBUTING.md LINK: "docs/RELEASING-old.md" is linked but not in the repository
$ echo $?
1

$ sed -i 's|https://github.com/vyncint/termlens/blob/main/docs/DESIGN.md|docs/DESIGN.md|' README.md
$ .github/scripts/check-readme-links.sh
README LINK: relative target "docs/DESIGN.md" — crates.io rewrites it against
  crates/termlens/, where it does not exist. Use https://github.com/vyncint/termlens/blob/main/docs/DESIGN.md
$ echo $?
1
```

With the three files restored, the gate and the docs-consistency gate CI already
runs both stay green:

```
$ .github/scripts/check-readme-links.sh README.md CONTRIBUTING.md
README.md: 24 distinct link target(s)
CONTRIBUTING.md: 15 distinct link target(s)
markdown links: 39 target(s) across 2 file(s), every in-repo target present   # exit 0

$ .github/scripts/check-ci-gates-listed.sh
every CI gate is listed in CONTRIBUTING.md §1 (19 commands, zizmor==1.29.0)   # exit 0

$ .github/scripts/check-dco.sh main..HEAD && .github/scripts/check-no-ai-attribution.sh main..HEAD
check-dco: OK
check-no-ai-attribution: OK
```

**What I did not do**

* No `CHANGELOG.md` entry: nothing here is user-facing, it is the contributor
  gate itself. Say the word if you want one anyway.
* The file is still called `check-readme-links.sh` although it now reads two
  files. A rename touches `ci.yml`, `CONTRIBUTING.md` and a `CHANGELOG.md`
  entry that names the old spelling, so I kept it out of this diff — happy to
  do it in a follow-up if you prefer the honest name.
* The org list is data in the script (`known_repos`), not a network call. If a
  fifth repository ever lands, the gate fails on its links until the list grows
  by one word, which seemed the right trade against a flaky external check.

Closes #352


vyncint (approved):
Thanks — this is an unusually complete piece of work, and it closes #352 on every point.

I checked each Done-when item against your branch rather than taking the evidence in the description on trust:

- A misspelled `vyncint/temlens` link in `CONTRIBUTING.md` fails, and the selftest pins it.
- A renamed in-repo target linked from `CONTRIBUTING.md` fails.
- The absolute-only rule still applies to `README.md` alone, and CONTRIBUTING's six legitimate relative links still pass.
- The reason the two rules differ is in the script header, where the next person editing either half will actually read it.

The selftest is the part I liked most. Proving the gate can go red before trusting it to be green is what `tools/semver-gate-selftest/` established, and following that shape without being asked is the right instinct. Running it in `gates-listed` is the right home for it too, since it is shell-only and single-OS.

### One thing I pushed a commit for

`known_repos` held four names and the comment called them "the four that share this contributor pattern". That was true when the pattern was written down and is not any more: oxmera and oxidelake have since joined it, and `docs/DESIGN.md` already links `vyncint/termlens-demo` for the coverage study. So the gate reported a real, already-linked repository as `names no repository in the org`.

It costs nothing today, because the gate reads two files and none of those three appear in them. It would have cost the next person, though, and precisely because of what you built: the file list is a parameter now, so pointing the gate at `docs/DESIGN.md` is the obvious next step, and it would have failed on a correct link with a message asserting something false.

What I changed:

- The list gains `oxmera`, `oxidelake` and `termlens-demo`. It stays deliberately short rather than becoming every repository the account holds, for your reason: a list of everything waves through the typo the gate exists to catch.
- The message no longer claims the repository does not exist. It says the gate has not been told about it, and how to add one. A gate should not assert a falsehood while failing.
- The good fixture links `termlens-demo`, so the distinction between "typo" and "real but unlisted" is pinned instead of argued.

Your `temlens` fixture still fails, which is the thing that matters.

Two notes on what you said you did not do, and I agree with both. No CHANGELOG entry is right, since nothing here is user-facing. And leaving the rename for a follow-up is right too, since it touches `ci.yml`, `CONTRIBUTING.md` and a CHANGELOG entry naming the old spelling. If you want to do that follow-up, I will take it.

PR #406, opened 2026-09-16
docs: drop the decode test count from CONTRIBUTING §1

### LINKED ISSUE #387: CONTRIBUTING §1 says `decode` gates "sixteen tests"; it is eighteen
**Today** — the sentence that explains why `--all-features` is not optional
carries a count that has drifted:

```console
$ cargo test -p termlens -- --list | grep -c ': test$'
494
$ cargo test -p termlens --features decode -- --list | grep -c ': test$'
512
$ grep -n sixteen CONTRIBUTING.md
32:`--all-features` matters: `decode` is off by default, and without it sixteen
```

512 − 494 = 18. The two added since the sentence was written are the
`colours` timing case and the HLS-refusal case in the graphics tests.

**Why it is worth fixing** — §1 is the one place the repository explains to
a newcomer why the documented test command has `--all-features` in it, and
this is the half of that section nothing checks. A number that is visibly
wrong invites doubt about the rest of a page whose whole purpose is to be
trusted literally.

**Fix** — in `CONTRIBUTING.md`, either correct the number or reword so the
sentence carries no count, which stops it drifting with the next `decode`
test added. Say which you chose; a maintainer may prefer the exact number
for its concreteness, in which case a one-line check could keep it honest.

**Done when**
- [ ] The sentence matches the measured difference, or names no count.
- [ ] `.github/scripts/check-ci-gates-listed.sh` still passes.

---

maskjelly:
Closes #387.

**What changed.** §1 of CONTRIBUTING explained that `--all-features` matters because `decode` is off by default, and said sixteen tests never build without it. The measured difference is eighteen today — `cargo test -p termlens -- --list | grep -c ": test\$"` gives 494, and 512 with `--features decode` — so the number had drifted by two as the `decode` tests grew.

**Which option I chose, and why.** The sentence names no count now:

> `--all-features` matters: `decode` is off by default, and without it the `decode` tests never build.

I chose that over correcting sixteen to eighteen because a number here tracks every `decode` test the suite gains, and this one had already gone stale unnoticed. The alternative — keep the number and add a check that keeps it honest — would spend a cargo invocation in CI on a documentation figure whose wording does not depend on it. The paragraph still says why the flag is not optional, which was its job.

**Not done, deliberately.** No `decode`-count gate was added, and the "eighteen" figure is recorded only in this description, so nothing else needs to move when the count changes again.

**Gates run locally:** `check-ci-gates-listed.sh` and `check-readme-links.sh README.md CONTRIBUTING.md`, both green. Markdown-only change, so no test accompanies it; `cargo fmt`/`clippy` untouched.


vyncint (approved):
Thanks! Merging.

You took the second of the two options #387 offered — dropping the count rather than correcting it to eighteen — and that is the better one here. The number drifts with every `decode` test anyone adds, and nothing checks it, so a sentence that carries no count is a sentence that cannot go stale again. `gates-listed` still passes, which is what mattered for this file.

I pushed one commit on top, `68ffafc`: the rewording left a line at 82 columns where the prose in `CONTRIBUTING.md` wraps at 72–76, so I reflowed the paragraph. No wording change — your sentence is unchanged, only the line breaks moved.

Issue #384, opened 2026-09-16
Three links in `docs/LIMITATIONS.md` are 404s, and the link gate cannot see them: it resolves relative links against the repository root, not the linking file


vyncint:
**Today** — LIMITATIONS links its siblings with a path that is correct from
the repository root and wrong from where the file actually lives:

```console
$ grep -n '](docs/' docs/*.md
docs/LIMITATIONS.md:68:  [docs/DESIGN.md](docs/DESIGN.md) §2.
docs/LIMITATIONS.md:149:  [docs/STABILITY.md](docs/STABILITY.md).
docs/LIMITATIONS.md:157:  [docs/DESIGN.md](docs/DESIGN.md).
```

From `docs/LIMITATIONS.md` those resolve to `docs/docs/DESIGN.md`, which
does not exist. Pointing the gate straight at the file does not catch it,
because the gate resolves every relative link against the repository root:

```console
$ .github/scripts/check-readme-links.sh docs/LIMITATIONS.md
docs/LIMITATIONS.md: 3 distinct link target(s)
markdown links: 3 target(s) across 1 file(s), every in-repo target present
$ echo $?
0
```

and `ci.yml` passes it only `README.md CONTRIBUTING.md`, so the file is
never checked in the first place. The self-test bakes the wrong rule in
too: its good fixture links `docs/RELEASING.md` from a file three
directories down and passes.

**Why it is worth fixing** — LIMITATIONS is where the README sends every
reader for "the full one, with the reason behind each entry", and three of
its cross-references are dead. [#352](https://github.com/vyncint/termlens/issues/352)
built this gate to stop exactly this, and it reports green.

**Fix** — three parts, all small:

1. Resolve a relative link against the linking file's directory in
   `.github/scripts/check-readme-links.sh`.
2. Move the self-test fixture so it proves the new rule, and add a case for
   a link that resolves from the root but not from the file.
3. Fix the three links, and pass `docs/*.md` and `skills/termlens/SKILL.md`
   to the gate in `ci.yml` and in CONTRIBUTING §1.

**Done when**
- [ ] The gate fails on today's `docs/LIMITATIONS.md` and passes after the
      links are fixed.
- [ ] `tools/link-gate-selftest/run.sh` covers the root-versus-file case.
- [ ] `ci.yml` checks the docs and the skill, and CONTRIBUTING §1 lists the
      same command.

---

*A good first contribution: one file, no architecture decisions. Start with
[...]

PR #407, opened 2026-09-16
fix(ci): check the pull-request template's links

### LINKED ISSUE #369: Both CONTRIBUTING links in the pull-request template are 404s in every pull request that uses it
**Today** — `.github/PULL_REQUEST_TEMPLATE.md` links the two checklist
items a contributor is most likely to click as `../CONTRIBUTING.md`. That
is right for a file living in `.github/`, and wrong for a pull-request
body, which GitHub renders relative to the pull request's own URL. The
href is emitted verbatim:

```console
$ curl -s -L https://github.com/vyncint/termlens/pull/351 |
>   grep -o 'href="[^"]*CONTRIBUTING[^"]*"' | sort -u
href="../CONTRIBUTING.md"
```

From `/vyncint/termlens/pull/351` that resolves to
`https://github.com/vyncint/termlens/CONTRIBUTING.md`:

```console
$ curl -s -o /dev/null -w '%{http_code}\n' -L https://github.com/vyncint/termlens/CONTRIBUTING.md
404
$ curl -s -o /dev/null -w '%{http_code}\n' -L https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md
200
```

**Why it is worth fixing** — the two dead links are on the two rules this
project actually blocks a merge over: the DCO sign-off and the no-AI-attribution
policy. A contributor who has just been told their commits are unsigned
clicks the link for the fix and lands on a 404. It is the same class as
[#341](https://github.com/vyncint/termlens/issues/341) — a relative link
correct in the repository and broken where the file is really read.

**Fix** — make both links absolute, with the section anchors while you are
there, e.g.
`https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md#5-developer-certificate-of-origin-dco`
and the `#6-…` anchor beside it. Check the anchors against the headings in
`CONTRIBUTING.md` rather than assuming the slug.

**Done when**
- [ ] Neither checklist item links a relative path, and both anchors land
      on the right section.
- [ ] `.github/ISSUE_TEMPLATE/*.yml` is checked for the same shape at the
      same time.
- [ ] `.github/scripts/check-readme-links.sh` — which already takes its
      files as arguments — is considered for the template: it resolves
      in-repo links, so a relative link there should either be allowed
      deliberately or caught. Say in the PR which you chose and why.

---

maskjelly:
Closes #369.

**What changed.** Both checklist links in `.github/PULL_REQUEST_TEMPLATE.md` were `../CONTRIBUTING.md`. That is correct for a file living in `.github/` and wrong where the template is actually read: GitHub emits the href verbatim into every pull request body, where it resolves against the pull request's URL. From `/vyncint/termlens/pull/351` that is `/vyncint/termlens/CONTRIBUTING.md` — a 404 — on the two rules this repository blocks a merge over, the DCO sign-off and the AI-attribution policy.

Both links are absolute now, with the anchors verified against the rendered page rather than guessed: `#5-developer-certificate-of-origin-dco` and `#6-ai-tooling-policy`.

**Which option I chose for the gate, and why.** The issue asked for the template to be considered for `check-readme-links.sh`, with the relative form either allowed deliberately or caught. I chose **caught**: the template gets the same absolute-only rule as the packaged README, because it has the same failure mode from a different URL. A relative target in a pull-request body can never resolve — there is no directory to be relative to — so allowing one deliberately is not a shape worth permitting. The script resolves relative links against the repository root for files that may use them (`CONTRIBUTING.md`), which is why this needed the rule rather than just the invocation.

**Tests.** `tools/link-gate-selftest/run.sh` gains a `pr-template-relative/` fixture and asserts the gate fails on it (it does: exit 1), and the "repository's own files" assertion now includes the template so the fixed links are checked too. 7 expectations, all green.

**Also checked, as the issue asked.** `.github/ISSUE_TEMPLATE/` for the same shape: `bug_report.yml` and `feature_request.yml` carry no links, and `config.yml`'s only link is already absolute, so nothing there needed changing.

**Not done, deliberately.** The gate is not extended over `AGENTS.md`/`CLAUDE.md`/`docs/`, which the issue did not ask for; `CONTRIBUTING.md` was the only non-README file in the gate's scope before this PR.

**Gates run locally.** `tools/link-gate-selftest/run.sh`, `check-readme-links.sh README.md CONTRIBUTING.md .github/PULL_REQUEST_TEMPLATE.md`, `check-ci-gates-listed.sh` — all green. Markdown, one shell script and a workflow step; no Rust touched.


vyncint (approved):
Thanks! Merging.

The fix itself is right, and it bites where it counts: I reverted the two template links back to `../CONTRIBUTING.md` on your branch and CI went red with your message, naming the reason rather than just "missing target". Both anchors resolve to the real headings, and you checked `.github/ISSUE_TEMPLATE/*.yml` rather than assuming — `config.yml`'s link was already absolute.

I especially liked that you made the "allow deliberately or catch" choice explicitly and argued it from the failure mode: a pull-request body has no directory for a relative link to be relative *to*. That is the reasoning the issue asked for, and it is the part that will still make sense to someone reading the script in a year.

I pushed one commit, `68e6e7d`, because the selftest fixture did not isolate the new rule. It linked `../CONTRIBUTING.md`, which the gate already rejected before your branch — relative targets resolve against the repository root, and `$root/../CONTRIBUTING.md` does not exist. Since `expect()` compares exit status only, a fixture that goes red for the *old* reason satisfies the new expectation. Reverting `check-readme-links.sh` and keeping your fixture left the selftest 7/7 green:

```
link gate selftest: 7 expectation(s), the gate fails when it should
```

Changing the fixture to link `CONTRIBUTING.md` fixes it: that target resolves from the root, so the old rule accepts it and only the new absolute-only rule can reject it. With the script reverted it now fails, which is the whole point of a gate selftest:

```
FAIL  PULL_REQUEST_TEMPLATE.md: a relative target fails (#369)   exit 0, expected nonzero
```

This is the same shape as the existing `readme-relative/` fixture, which links `docs/DESIGN.md` for exactly that reason.

Worth saying: #384 is a separate open issue and this PR neither fixes nor regresses it. The gate still resolves relative links against the repository root, so `docs/LIMITATIONS.md`'s three broken links stay invisible for now.

Issue #298, opened 2026-09-08
A hidden cursor makes a saved-screen round trip report a difference with no visible change


vyncint:
`Screen::parse` promises that parsing a with-styles snapshot diffs empty against the original. That fails for a hidden cursor away from (0,0): the text format omits its position, but `ScreenDiff::is_empty()` compares the position anyway.

### termlens version

Reproduced on `0.10.0`, tag SHA `0a4b0cec2d7bcc948412b848b87c6fc05b0222d4`, and independently in a consumer using the published registry dependency `termlens = "=0.10.0"`.

The affected library source is unchanged at observed main commit `daff78f25349c84facc143c062daad9d0acbe51a` (source comparison; the reproduction below was run against the release).

### Operating system

macOS 26.6.2 (25G83), ARM64; Rust stable 1.97.1; debug test profile.

### TERM value

`xterm-256color` (TerminalBuilder default).

### Minimal reproduction

In a checkout of tag `v0.10.0`, save this complete integration test as `crates/termlens/tests/repro_diff_hidden_cursor.rs`. It uses the existing public `tests/common` helper and `emit` fixture; no application-specific binary is needed. Run:

```sh
cargo test -p termlens --all-features --locked --test repro_diff_hidden_cursor -- --nocapture
```

```rust
use std::time::Duration;
use termlens::{Screen, Terminal};

mod common;

fn capture(text: &str) -> termlens::Result<Screen> {
    let mut t = common::spawn_emit(
        Terminal::builder()
            .size(16, 4)
            .timeout(Duration::from_secs(5)),
        &[text],
    )?;
    t.wait_exit()?;
    Ok(t.screen())
}

#[test]
fn hidden_cursor_round_trip_is_the_same_picture() -> termlens::Result<()> {
    let screen = capture("hello\x1b[2;3H\x1b[?25l")?;
    assert_eq!(screen.cursor(), (1, 2, false));
    let parsed = Screen::parse(&screen.with_styles().to_string())?;
    assert!(screen.diff(&parsed).is_empty(), "{}", screen.diff(&parsed));
    Ok(())
}
```

### Expected vs actual

**Expected:** The parsed snapshot should diff empty against the original when the cells, size and visible cursor state are the same, as the parse rustdoc promises.

**Actual:** The original cursor is `(1, 2, false)`, while the parsed cursor is `(0, 0, false)`. `diff.is_empty()` is false and its rendering names no visible change, only `cursor: hidden` and four unchanged rows.

```text
[...]

PR #408, opened 2026-09-16
fix(serde): refuse a cursor one column past the right edge

### LINKED ISSUE #375: The JSON reader accepts a cursor one column past the right edge, so `termlens render` writes a screen its own parser refuses
**Today** — the cursor bound check in the `serde` reader is `>` where the
row check beside it is effectively `>=`, so `col == cols` gets through.
On a 4x2 screen:

```
cursor.col=4: Ok((0, 4, true))     ← accepted, and there is no column 4
cursor.col=5: Err("cursor 0,5 is outside a 4x2 screen")
cursor.row=2: Err("cursor 2,0 is outside a 4x2 screen")
```

The CLI then converts a file it accepts into a file it rejects:

```console
$ termlens render --text badcursor.json
size: 4x2  cursor: 0,4
ab
cd
$ echo $?
0
$ termlens render --text badcursor.json > rt.txt
$ termlens render --text rt.txt
termlens: rt.txt: could not parse a saved screen: line 1: cursor 0,4 is outside the 4x2 grid
$ echo $?
2
```

**Why it is worth fixing** — `docs/STABILITY.md` promises the JSON as a
persisted format that `termlens diff` and `render` read back, and
`serde_impl.rs`'s own header says a hand-edited file "is an error and never
a screen that panics on its first `cell()`". One off-by-one lets JSON
express a cursor the text format cannot, so a round trip between the two
promised formats fails on a file that passed validation.

**Fix** — in `crates/termlens/src/screen/serde_impl.rs`, make the column
bound `>=` like the row bound, guarding the zero case the same way the row
check does.

**Done when**
- [ ] `cursor.col == cols` deserialises to `Err` with the existing
      `cursor R,C is outside a WxH screen` message.
- [ ] A case in `crates/termlens/tests/export.rs` pins both `col == cols`
      and `row == rows` as errors — only the row form is covered today.
- [ ] Every file in `crates/termlens/tests/compat/` still parses, and
      `cargo test --workspace --all-features` passes.

---

maskjelly:
Closes #375.

**What changed.** The JSON reader's column bound was `>` where the row bound beside it was `>=`, so `col == cols` — a column that does not exist — passed validation. On a 4x2 screen `cursor.col=4` deserialised `Ok((0, 4, true))` while `col=5` and `row=2` were refused, and `termlens render --text` then converted a file it had accepted into a file its own parser rejected (`cursor: 0,4 is outside a 4x2 grid`). One off-by-one let JSON express a cursor the text format cannot, so a round trip between the two promised formats failed on a file that had passed validation.

The column bound is `>=` now, guarding zero the way the row check does. `tests/export.rs` pins both `col == cols` and `row == rows` as errors with the existing `cursor R,C is outside a WxH screen` message, and the last real cell still reads.

**The one accommodation outside the issue, and why it cannot weaken anything.** The ratatui fixture's expected-screen builder reads the PTY screen back through JSON. Ratatui leaves its hidden cursor at the pending-wrap position (one past the last column), which the reader now refuses, so that builder clamps the position — only when the cursor is not visible, where `same_cursor` does not compare it (#298). A visible cursor keeps its position, so a real disagreement would still fail the test.

**What I did not do, and a caveat worth knowing.** The writers still emit the pending-wrap position. `inspect` on a 4x2 screen after `printf abcd` writes `cursor: 0,4`, which `Screen::parse` refuses today (pre-existing, text side) — and after this change the JSON side has the same property: a live screen with `col == cols` serialises to JSON its own reader refuses. I left the writers alone because the issue scopes the reader and asks for no restructuring, but if you would prefer the round trip to hold for pending-wrap screens, clamping at the writer (or normalising at snapshot time) is the other half — happy to add it here or in a follow-up. No file in `tests/compat/` holds a pending-wrap cursor.

**Evidence.** On a 4x2 screen: col=4 and row=2 → `Err`, col=5 → `Err`, col=3 → `Ok`. The new test fails on the unfixed reader at the first case and passes after. `cargo test --workspace --all-features` green (48 `test result: ok` blocks, none failed), fmt and clippy clean.


vyncint (approved):
Thanks! Merging — and the caveat section in your PR body is the reason this went in safely.

You wrote down the residual defect before anyone found it: the reader was tightened while the *writers* still emit `col == cols`, so a live screen filled to its right margin would serialise to JSON its own reader refuses. That was real. I measured it on a 4x2 PTY: on main the JSON round trip passed and the text one failed; with your change alone, both failed. Tightening the reader by itself would have traded a JSON-only inconsistency for a break of the `docs/STABILITY.md` promise that JSON written by 0.10 or later stays readable.

Rather than ask you to clamp the writers, I merged #430 first. It clamps the pending-wrap cursor at the emulator, so no screen carries that column at all — the same 4x2 screen now reports `(0, 3, true)` and both round trips pass with your reader in place. That is why this one waited.

That also made your `fidelity.rs` workaround unnecessary, so I pushed `e598e06` removing it; the fidelity test passes without it. Your comment there was correct about why a *visible* cursor must not be clamped, and I checked `same_cursor` before removing it — two hidden cursors compare equal regardless of position, so the clamp could not have masked a real disagreement either way.

The bound itself is right: `col == cols` now fails, `col == cols - 1` still passes, and the row bound is untouched. Your test pins both axes, where only the row form was covered before.

PR #409, opened 2026-09-16
fix(screen): name the real caller in a reversed-range panic

### LINKED ISSUE #379: `mask_rect` panics with a message naming `rect_text`, a method the caller never called
**Today** — both methods share a range-clamping helper that hard-codes one
name into the panic:

```console
thread 'main' panicked at crates/termlens/src/screen.rs:2194:
rect_text: column range starts at 3 but ends at 0
```

That is from `screen.mask_rect(3..0, ..)`. `mask_rect(.., 2..1)` says
`rect_text: row range starts at 2 but ends at 1`. `rect_text(3..0, ..)`
produces the identical text, so nothing in the panic distinguishes the two
call sites — in a test file that may well use both.

**Why it is worth fixing** — this panic exists precisely to catch a
transposed `(col, row)` pair, which the source explains at length as the
mistake worth failing loudly on. Sending the reader to the wrong method
name costs them the one clue the panic was built to give. `mask_rect`'s own
`# Panics` section says it panics *as* `rect_text` does, not that it claims
to be `rect_text`.

**Fix** — give the helper in `crates/termlens/src/screen.rs` a caller-name
parameter and pass `"rect_text"` and `"mask_rect"` from the two call sites.
Panic prose is explicitly unpromised in `docs/STABILITY.md`, so nothing
breaks.

**Done when**
- [ ] `mask_rect(3..0, ..)` panics with `mask_rect: column range starts at
      3 but ends at 0`.
- [ ] The existing reversed-range unit tests gain a `mask_rect` twin with
      `#[should_panic(expected = "mask_rect: ")]`.

---

maskjelly:
Closes #379.

**What changed.** `mask_rect` and `rect_text` share `clamp_range`, which hard-coded `rect_text: ` into its panic. `mask_rect(3..0, ..)` claimed to be `rect_text` — and this panic exists precisely to catch the transposed `(col, row)` pair, so the wrong name cost the reader the one clue it was built to give. The helper takes the caller's name now and each call site passes its own.

**Tests.** The two existing `rect_text` expectations gained the `rect_text: ` prefix, and the `mask_rect` twins pin `mask_rect: ` on both axes. On the unfixed helper the twins fail with `expected substring: "mask_rect: ..."` against the exact pre-fix panic text; after the fix all four pass.

**Not done, deliberately.** No CHANGELOG entry: panic prose is unpromised in `docs/STABILITY.md`, so there is no user-facing change to announce. The two `rect_text` expectation strings were strengthened slightly past the literal Done-when so a regression where both call sites pass the same name is caught; happy to drop that half if you want the letter of the issue only.

**Checks.** `cargo test --workspace --all-features` green, fmt and clippy clean.


vyncint (approved):
Thanks! Merging.

The fix is right and, more importantly, the tests bite: I reverted the format string to the hard-coded name and exactly the two new tests failed with `panic did not contain expected string`, while the two `rect_text` ones kept passing. That is the shape you want — a change that reintroduces the bug fails the new tests and nothing else.

Strengthening the two *existing* `rect_text` expectations to include the `rect_text:` prefix was beyond the letter of the issue and is the better instinct. Without it, both call sites passing the same name would still slip through. And you placed the new tests before the following doc comment rather than between it and its function, which is a mistake I have already had to fix once in this batch.

I pushed one small commit, `9ff2c88`, giving the pair a doc comment saying what they defend against — that the panic names the method the caller actually called. Every neighbouring test has one, and here it is the whole point of the change.

On the CHANGELOG: I agree with your reasoning that panic prose is unpromised in `docs/STABILITY.md`, so I have not asked for an entry.

PR #410, opened 2026-09-16
fix(screen): render blink in the SVG and HTML export

### LINKED ISSUE #378: Blink is the one `Style` attribute of eight that reaches neither `to_svg` nor `to_html`, so a blinking cell renders identically to a steady one
**Today** — holding everything else equal on a `fg=1` run and adding each
attribute in turn, then comparing the rendering to the unstyled one:

| attribute | SVG differs | HTML differs | ANSI differs |
| --- | :-: | :-: | :-: |
| bold | ✔ | ✔ | ✔ |
| dim | ✔ | ✔ | ✔ |
| italic | ✔ | ✔ | ✔ |
| underline | ✔ | ✔ | ✔ |
| **blink** | **✗** | **✗** | ✔ |
| reverse | ✔ | ✔ | ✔ |
| conceal | ✔ | ✔ | ✔ |
| strikethrough | ✔ | ✔ | ✔ |

`to_ansi` carries it as `SGR 5` and the `styles:` block names it. Only the
two renderings made for attaching to a pull request drop it, and neither
rustdoc mentions the omission while both enumerate their other special
cases (conceal renders as blanks, dim as opacity).

**Why it is worth fixing** — blink is the attribute the shadow parser
exists to recover: `docs/STABILITY.md` §2 keeps a second parser over the
whole stream because `vt100` drops blink, conceal and strikethrough, and
blink is the one that kept `vt100` from being replaced. A blink regression
is therefore invisible in exactly the artefact a reviewer looks at.

**Fix** — in `crates/termlens/src/screen/render.rs`, pick one and say which
in the PR:

1. **Render it.** HTML: one `@keyframes` rule in the `<style>` the `<pre>`
   already carries, and an `animation:` on blinking spans. SVG: an
   `<animate attributeName="opacity">` child on the blinking `<text>`.
2. **Document it.** State the omission in both rustdocs and in
   `docs/LIMITATIONS.md` under "Out-of-band state and styles", beside the
   overline and double-underline entry.

Either is one file plus a test. `docs/STABILITY.md` does not promise the
renderings' exact bytes, so (1) breaks nothing.

**Done when**
- [ ] Two screens differing only in `blink` render differently, **or** both
      rustdocs and LIMITATIONS.md name blink as not rendered.
- [ ] A test in `crates/termlens/tests/export.rs` covers all eight
      attributes in one table, so the next attribute cannot be dropped
      silently.

---

maskjelly:
Closes #378.

**Which option, and why.** Option 1, render it. The issue lists it first, and the artefact a reviewer looks at is where blink vanishing matters most. There was no genuine blocker; the omission was plain.

**What the renderings do.** SVG: an `<animate attributeName="opacity">` child on a blinking `<text>`, `discrete` so it toggles rather than fades. A dim blink carries `0.6` in its values because an SMIL animation outranks the attribute it animates, or it would animate back to full opacity. HTML: a `termlens-blink` `@keyframes` rule in a `<style>` element inside the `<pre>`, and `animation:` on blinking spans. The `<pre>` carried no `<style>` element before — a `style` attribute cannot hold `@keyframes` — so the element is written only when a run blinks; a screen without blink renders byte for byte as before, which is why no snapshot moved. The animated property is the glyph's colour, not opacity, so a dim blinking run keeps its dim and the run's background keeps painting, as a terminal's blink hides the glyph and not the cell.

**Test.** `tests/export.rs` gains one table over all eight attributes in all three renderings: each must differ from the unstyled screen and carry its rendering's own marker. It fails on today's code at exactly the blink row (`assertion left != right: SVG drops blink`), passes after.

**Honest gaps.** (1) The `<style>` element inside `<pre>` is not strictly valid per the HTML content model, though it works in browsers; it is where the issue suggested the keyframes live. A sanitizer that strips `<style>` (GitHub step summaries do sanitize) drops the animation and lands back on today's steady glyph — no regression, but no blink there either; the SVG is the artefact actually attached as an image. (2) 1 s cycle and a hard 50% toggle are deliberate choices, not measured against a particular terminal. (3) `docs/LIMITATIONS.md` is untouched — option 2's route was not taken.

**Checks.** `cargo test --workspace --all-features` green; fmt and clippy clean; no `.snap.new` anywhere.


vyncint (approved):
Thanks! Merging.

You picked the harder of the two options the issue offered and the dim interaction is the part that shows it was done properly: an SMIL animation outranks the attribute it animates, so a dim blink would flash back to full brightness unless the animation carries dim's own 0.6 in its values. You handled that and pinned it with its own assertion. Choosing `color` in HTML but `opacity` in SVG, so the cell background keeps painting the way a real terminal's does, is the correct reading of what blink means.

I diffed both renderings for twelve style combinations against main: only blink-bearing screens differ, the other seven attributes are byte-identical, and every SVG still parses as XML. Reverting the three render hunks fails the new test at exactly the blink row. No snapshot needed updating, because none of the committed ones contains a blinking cell.

I pushed `c8666ac` for accessibility. Both animations ran `indefinite`/`infinite` with no way to stop them, which is WCAG 2.2.2 — and this crate already cares, since 0.11.1 gave the SVG an accessible name for the same reason. The HTML rendering now carries a `prefers-reduced-motion` rule beside its keyframes, matching on the inline style since that is where the `animation` lands.

The SVG cannot be fixed the same way: SMIL is not reachable from CSS and `<animate>` has no media-query form. Rather than leave that unsaid, both rustdocs and `docs/LIMITATIONS.md` now state it and point a reader who needs the motion stopped at the HTML rendering.

You were also right that the issue text was wrong about the `<pre>` carrying a `<style>` element — it carries a `style` attribute. Good catch.

PR #411, opened 2026-09-16
fix(emu): keep scrollback across a hard reset

### LINKED ISSUE #391: A hard reset makes `scrollback_text()` drop rows out of the middle, and which rows depends on how the output was chunked
**Today** — `ESC c` (RIS) rebuilds the backend's screen and empties its
history, but termlens's high-water mark for "rows already captured" is
never reset. Rows that scroll after the reset are compared against a stale
mark and skipped. On a 20x4 grid with `scrollback(20)`, printing six lines,
resetting, then printing more:

```
six lines, reset, six lines   grid="D|E|ZZ|"   scrollback="1|2|3"
six lines, reset, nine lines  grid="G|H|ZZ|"   scrollback="1|2|3|A|B|C|D|E|F"   (correct)
one write, same bytes         grid="G|H|ZZ|"   scrollback="1|2|C|D|E|F"
```

The first loses `A B C` outright. The third loses `3`, `A` and `B` and
interleaves what is left. The second is correct — the difference is only
how the PTY happened to chunk the reads.

**Why it is worth fixing** — silent history loss, and chunk-dependent, so a
test that asserts on scrolled-off output near a `reset`, a `tput reset` or
a TUI's exit sequence flakes rather than fails. A flake in a test harness
is the most expensive kind of bug it can have.

**Fix** — in `crates/termlens/src/emu/vt100.rs`, the scroll-capture routine
assumes the backend's history only ever grows. Before the growth test, note
that a shorter history means the screen was rebuilt and reset the mark to
the start. Record RIS in the comment block that states the invariant, since
that block currently says the assumption holds unconditionally.

**Done when**
- [ ] All three shapes above produce the full history in order.
- [ ] A test in `crates/termlens/tests/scrollback.rs` scrolls, sends `ESC c`
      in a later write, scrolls again, and asserts no row is missing.
- [ ] The comment in `emu/vt100.rs` names RIS as the case that breaks
      monotonicity.

---

maskjelly:
Closes #391.

**What changed.** `ESC c` (RIS) rebuilds vt100's screen and empties its history, but termlens' high-water mark of rows already captured survived it. Rows scrolling after the reset were compared against a stale mark and skipped, and which rows went missing depended on chunking: with six lines, a reset and six more on a 20x4 grid, one chunking produced the correct `1|2|3|A|B|C` while another produced `1|2|3`, and a third lost `3`, `A` and `B` and interleaved the rest. Rows that scrolled *before* the reset within the same chunk were never captured at all — by the time the parser saw the chunk, RIS had wiped them.

**The fix is three pieces, because the mark reset alone is not enough.** It handles rows that scroll after the reset, but the pre-reset rows in the same chunk are gone before any capture runs, so:

1. `emu/seq.rs` emits `SeqEvent::HardReset` when `ESC c` completes — deliberately not for `CSI c` (device attributes), and a test pins that.
2. `emu/vt100.rs` splits its feed at that byte: the bytes before it are fed first, so their capture takes the pre-reset rows while the backend still holds them; the reset is then fed alone, and the capture that follows sees the emptied history, which is how the mark learns to start over.
3. The capture routine handles a history shorter than the mark before the growth test, and its invariant comment names RIS as the case that breaks monotonicity; `docs/DESIGN.md` carries the same qualification.

**Tests.** `tests/scrollback.rs` pins the reset in a write of its own, merged with the rows that follow it, and the whole stream in one write, asserting history order and that no row is missing; `emu/vt100.rs` pins all seven split points around the reset byte; the tracker test pins that `CSI c` does not fire the event. The repro harness produced the issue's three shapes before the fix and the full ordered history after. The new integration test was run 20x and its file 10x continuously.

**Not done / observed.** (1) The new integration test spawns the emit fixture, so CONTRIBUTING §3's stress requirement applies to it; I cannot dispatch workflows from a fork — could you run `stress.yml` on this branch, or say if enabling Actions on my fork and dispatching there is acceptable to you. (2) An adjacent bug is left alone deliberately: entering and leaving the alternate screen duplicates main-screen history (`captured` is zeroed while the alt grid has no history and the exit re-copies from 0). I reproduced it on unmodified pre-fix logic; it is not #391 and I did not test a fix. (3) CHANGELOG updated under Unreleased.

**Checks.** `cargo test --workspace --all-features` green (48 `test result: ok` blocks, none failed); fmt and clippy clean.


vyncint (approved):
Thanks! Merging.

The key insight here is yours and it is not obvious: resetting the mark alone cannot work, because by the time any capture runs the pre-reset rows are already gone from vt100's history. Splitting the feed at the reset byte — so the rows scrolled before it are captured while the backend still holds them — is the fix, and it mirrors `apply_tabs`'s existing shape rather than inventing a new one.

The unit test pinning all seven split points around the reset byte is the right instrument. This bug was chunk-dependent, so a test that fixes one chunking would have proved almost nothing; yours makes the chunk-dependence itself assertable.

I pushed `547595c` removing the other half. The `if len < self.captured { self.captured = 0; }` guard is unreachable once the feed split is in place: with the split, the capture after the reset always sees an empty history, and the mark re-syncs through the assignment already at the end of the function. I removed it and nothing failed — the whole library suite and the scrollback tests pass without it. Worse, in the one state where it *could* fire with a non-empty history it would set the copy start to zero and re-copy rows already held, duplicating them.

The same commit corrects `docs/DESIGN.md`, which credited that dead branch with the fix. Both the comment and the spec now name the feed split. I would rather the spec be right than the diff be small, and you were right that touching DESIGN.md at all was in scope — the issue asked for the invariant to record RIS.

I dispatched `stress.yml` on a branch carrying this and the rest of the batch, since a fork cannot: 15 legs green across all three platforms.

Your note about alternate-screen history duplication is real and pre-existing — I reproduced it identically on main — so it is not this PR's to fix. Worth its own issue.

Issue #383, opened 2026-09-16
`gates-listed` only extracts `cargo` steps, so the ten script gates are unchecked — and two of them are missing from CONTRIBUTING §1 right now


vyncint:
**Today** — CONTRIBUTING §1 promises that the `gates-listed` job "fails
when the two disagree, so the bug cannot sit here unnoticed again". Its
extractor matches `cargo` lines only:

```bash
ci_cmds="$(sed -n 's/^[[:space:]]*- run: \(cargo .*\)$/\1/p' "$ci" …)"
```

Ten gates in `ci.yml` are scripts rather than `cargo`, and two of them are
absent from §1's runnable block. The gate reports success anyway:

```console
$ bash .github/scripts/check-ci-gates-listed.sh
every CI gate is listed in CONTRIBUTING.md §1 (19 commands, zizmor==1.29.0)
$ echo $?
0
```

```
in-§1-block  check-candidate-statement.sh   NOT-IN-BLOCK check-ci-gates-listed.sh
in-§1-block  check-cli-contract.sh          NOT-IN-BLOCK check-skill-snippets.sh
in-§1-block  check-emit-steps.sh            in-§1-block  check-semver.sh
in-§1-block  check-feature-isolation.sh     in-§1-block  link-gate-selftest/run.sh
in-§1-block  check-readme-links.sh          in-§1-block  semver-gate-selftest/run.sh
```

**Why it is worth fixing** — a contributor who runs every line of §1 can
still be failed by the `gates-listed` and `skill` jobs, which is precisely
the bad first experience §1 says it exists to prevent. The success message
states the opposite, so nobody looks. It is
[#227](https://github.com/vyncint/termlens/issues/227)'s fault class
recurring past its own fix.

**Fix** — widen the extractor in
`.github/scripts/check-ci-gates-listed.sh` to also take `- run:` steps
beginning `.github/scripts/` or `tools/`, then add the two missing lines to
CONTRIBUTING §1. Make the success line report both counts, so a reader can
see the script gates are covered.

**Done when**
- [ ] The extractor picks up the script gates as well as the `cargo` ones.
- [ ] Deleting `check-skill-snippets.sh` from CONTRIBUTING §1 makes the
      gate go red — check this by hand before you trust it.
- [ ] `.github/scripts/check-ci-gates-listed.sh` passes on the fixed tree.

---

*A good first contribution: one file, no architecture decisions. Start with
[...]

PR #412, opened 2026-09-16
ci: check the report action's tag against the workspace version

### LINKED ISSUE #386: The report action's own README pins `report@v0.10.1`, two releases behind the other two copies, and RELEASING never says to bump them
**Today** — the same pin lives in three hand-maintained places, and the
copy on the action's own page is the stale one:

```console
$ grep -rn 'actions/report@' --include='*.md' --include='*.yml' .
README.md:232:- uses: vyncint/termlens/.github/actions/report@v0.11.1
.github/actions/report/README.md:13:- uses: vyncint/termlens/.github/actions/report@v0.10.1
skills/termlens/SKILL.md:552:… `uses: vyncint/termlens/.github/actions/report@v0.11.1`
```

The 0.11.1 release commit bumped two of the three and missed this one; so
did 0.11.0. `docs/RELEASING.md`'s "Cutting vX.Y.Z" checklist covers the
version bump, the CHANGELOG, the corpus and the semver baseline, and never
mentions the action pin.

**Why it is worth fixing** — the CHANGELOG tells consumers to "pin the
action at the same tag as the CLI it installs (`report@v0.11.1`); the flag
is new in 0.11.1", and the action's own README — the page a consumer lands
on — hands them a tag whose script predates `render --out`. Three copies of
one string with no gate is the class `check-candidate-statement.sh` already
exists for.

**Fix** — update `.github/actions/report/README.md` to the current tag, and
add a line to `docs/RELEASING.md`'s checklist naming all three files.
Better still, add a short gate beside `check-candidate-statement.sh` that
fails when a `report@` ref disagrees with `workspace.package.version`, and
list it in CONTRIBUTING §1 and `ci.yml`.

**Done when**
- [ ] All three `report@` refs match the workspace version.
- [ ] `docs/RELEASING.md` names them in the release checklist.
- [ ] If a gate is added, it is wired into `ci.yml` and CONTRIBUTING §1,
      and fails when one pin is edited by hand.

---

maskjelly:
Closes #386.

**What changed.** The `report@vX.Y.Z` tag is pinned in three hand-maintained files, and the workspace version is a fourth copy of the same number. The action's own README was still at `v0.10.1` — releases bumped two of the three and missed it twice. This PR:

1. Moves `.github/actions/report/README.md` to `v0.11.1`, so all three refs match `workspace.package.version`.
2. Names all three files in `docs/RELEASING.md` step 1, beside the flag gate.
3. Adds `check-report-pin.sh` — the "better still" from the issue — in CI's `docs` job beside `check-candidate-statement.sh`, and `tools/report-pin-gate-selftest/run.sh` in `gates-listed` beside the link gate's selftest, with `CONTRIBUTING` §1 listing both.

**What the gate checks.** Every `actions/report@vX.Y.Z` ref in the three files must equal `workspace.package.version`. It fails on a stale ref, on one stale ref among several, and on a listed file whose ref is gone entirely (a copy that quietly disappeared is drift too). The file list is a single default in the script, so the next file to carry the pin is a one-line addition. `REPORT_PIN_VERSION` exists only so the selftest can pin its fixtures to a version that does not move every release; CI never sets it, and the script header says so.

**How it is proven.** The selftest asserts the four failure shapes (matching passes; stale fails; absent fails; one stale among two fails) and — the assertion that keeps it honest — the repository's own three files against the real version. All five green locally, as are `check-candidate-statement.sh`, `check-readme-links.sh` and `check-ci-gates-listed.sh`.

**Not done, deliberately.** No CHANGELOG entry: no user-facing change, matching the link-gate precedent (#353). And the gate checks the three current files, not every `report@` string in the repository — a repo-wide grep would flag the historical pins CHANGELOG entries quote, which are correct as history. If you would rather it scan wider with an exclusion list, that is an easy change.

**One note on ordering.** The `CONTRIBUTING.md` §1 lines this adds sit in the same block as the link-gate invocation line that #407 touches; whichever of the two lands second, I will rebase it.


vyncint (approved):
Thanks! Merging as is.

I hand-edited the pin in each of the three real sites in turn and the gate went red on every one, naming the file and both versions:

```
::error::.github/actions/report/README.md pins `report@v0.10.1` but the workspace version is 0.11.1
```

Two things you did that the issue did not ask for and that make this gate worth having.

The `absent.md` fixture — a listed file that has *lost* its pin — covers a failure the issue never named, and it is the one that would let a silent deletion through. A gate that only checks the pins it finds is a gate that passes when someone deletes the line.

Pinning `REPORT_PIN_VERSION` to `9.9.9` in the fixtures keeps them from going stale at every release, while the last assertion still runs the real three files against the real workspace version. That is the difference between a selftest and a second thing to maintain.

You were also right to exclude the `report@v0.11.1` in `CHANGELOG.md`: it is historical prose under a released heading and must not move. Requiring the `actions/` prefix is the right way to say that.

One thing for a later PR, not this one: `check-ci-gates-listed.sh` extracts only `- run: cargo …` lines, so its green says nothing about either of your two new steps. You listed both in CONTRIBUTING §1 by hand and correctly, but nothing enforces it — that is #383, already open.

PR #413, opened 2026-09-16
docs: remove stale rust announcement draft

### LINKED ISSUE #388: `docs/announce-r-rust.md` still describes v0.8 and calls termlens Unix-only with Windows "planned"
**Today** — the file is an announcement draft, committed publicly, whose
own header says to delete it once posted. It has not been, and it has gone
stale in the way that matters most:

```console
$ grep -n 'v0.8\|Unix only\|Linux and macOS' docs/announce-r-rust.md
70:  CI runs the whole suite 100 times on Linux and macOS before anything merges
78:  Honest limitations (v0.8): Unix only for now (portable-pty supports ConPTY,
78:  so Windows is planned)
```

Windows has been a supported platform since 0.10 and is a required check on
every pull request — `docs/STABILITY.md` §1 is the decision and `README.md`
carries the platform table. The stress workflow has run `windows-latest`
since [#290](https://github.com/vyncint/termlens/issues/290), and it is
dispatch-only rather than a merge gate, so both halves of line 70 are also
wrong. The draft further predates `termlens-cli`, `Screen::diff`, the
`serde` feature, the agent skill and the stability candidate.

**Why it is worth fixing** — it is one copy-paste away from being posted
publicly with a false platform claim, and it is the only document in the
tree still describing 0.8. Whichever way it goes, the repository should not
carry a page that contradicts its own README.

**Fix** — pick one and say which:

1. **Delete it.** The header already sanctions this, and announcement prose
   is better written fresh at the moment of posting than maintained.
2. **Refresh it** against 0.11.1: the platform table, the CLI, the
   renderers, the skill, and the stability-candidate statement in the same
   words `check-candidate-statement.sh` enforces elsewhere.

**Done when**
- [ ] The file is gone, or no sentence in it contradicts `README.md` or
      `docs/STABILITY.md`.
- [ ] If it stays, `.github/scripts/check-readme-links.sh` is run over it.

---

Voyagerroc-Lab:
## Summary
- remove the stale r/rust announcement draft that still described v0.8 and Unix-only support
- update HANDOFF.md so it no longer points to the deleted draft
- leave future release announcements to be written from current project state

Closes #388

## Validation
- git diff --check
- verified no repository references to docs/announce-r-rust.md remain
- Cargo checks were unavailable because Cargo is not installed.


vyncint (approved):
Thanks! Merging as is.

You picked the delete option from #388, which is the one I would have picked too: announcement prose is better written fresh at the moment of posting than maintained against a moving crate, and the file's own header already sanctioned removing it.

What makes this complete rather than just a deletion is the second half. `docs/HANDOFF.md` pointed at the draft, and that pointer would have become a dangling reference that no gate would have caught — `check-readme-links.sh` only reads `README.md` and `CONTRIBUTING.md` today. You found it and rewrote the sentence so it still says something true about where announcements stand. I checked for any other reference to the file and there are none left in the tree.

PR #414, opened 2026-09-16
fix(graphics): reject zero kitty dimensions

### LINKED ISSUE #404: `decode()` accepts a kitty transmission declaring a zero dimension and returns an empty `Bitmap` instead of naming the contradiction
**Today** —

```
s=0, v=0   size() = Some((0, 0))   decode() = Ok(Bitmap 0x0)
s=8, v=0   size() = Some((8, 0))   decode() = Ok(Bitmap 8x0)
```

`pixel()` then returns `None` everywhere and the colour histogram is empty.
The length guard passes because the wanted byte count is zero, so any body
satisfies it, including an empty one.

**Why it is worth fixing** — a declared size of zero contradicts the
protocol: those keys are the pixel dimensions of a picture. The module's
stated position is that a contradiction should be an error naming the
reason rather than something a test could mistake for "the image was
empty", and an `Ok` of a zero-pixel bitmap is precisely that mistake in the
other direction — `decode()?` succeeds and the assertion after it silently
sees nothing. The sixel path already declines this: it falls back to the
painted extent when a declared dimension is zero.

**Fix** — one guard in the kitty decode path in
`crates/termlens/src/graphics.rs`, after the size is read, returning
`DecodeError::Malformed` with a message naming the zero dimension.

**Done when**
- [ ] `s=0,v=0`, `s=8,v=0` and `s=0,v=8` all return
      `Err(DecodeError::Malformed(_))`.
- [ ] A unit test covers all three, next to the existing
      every-refusal-names-its-reason case.

---

Rayan-and-beyond:
## What & why

Closes #404.

Kitty decoding could accept a declared width or height of zero because the required byte count collapsed to zero, which let an empty bitmap through as a successful decode. This rejects zero dimensions as malformed before the size calculation and adds coverage for `0x0`, `8x0`, and `0x8` declarations.

The changelog now records the corrected decode behavior under `[Unreleased]`.

## Validation

- `cargo test -p termlens --features decode --lib kitty_zero_dimensions_are_malformed -- --exact --nocapture`
- `cargo test -p termlens --features decode --lib graphics::tests`
- `cargo clippy -p termlens --lib --features decode -- -D warnings`
- `git diff HEAD^ --check`

I did not run the full workspace Clippy matrix locally because the available build host was disk-constrained; the focused decode Clippy check above is clean.

## Checklist


vyncint (approved):
Thanks! Merging.

The guard is right and it is in the right place — after the size is read and before anything touches the payload, so it runs ahead of the inflate bound rather than behind it. I checked it bites: removing the five lines makes your test fail with `a kitty transmission with a zero dimension was accepted: a=T,f=32,s=0,v=0`, which is what you want from a test like this. The three-case table covers both dimensions and the pair, and the message matches the one #404 asked for.

I pushed one commit, `9314039`, for something the compiler cannot see. The new test landed between the doc comment above `a_payload_cannot_choose_how_much_memory_a_decode_spends` and that function, so the comment describing the five allocation paths and why they are deliberately one table ended up documenting the zero-dimension case instead:

```
    /// One table on purpose, so a sixth path cannot appear without a line
    /// here saying what bounds it.
    #[cfg(feature = "decode")]
    #[test]
    fn kitty_zero_dimensions_are_malformed() {
```

An easy one to hit — the two tests look alike from the diff, and `cargo test` is perfectly happy either way. I moved yours down beside `every_refusal_names_its_reason`, which is where the refusals live and where #404 suggested it, and gave it a comment of its own. Your assertions are untouched.

I also rewrapped the CHANGELOG entry: it was one 130-column line where the file wraps at about 74, and I extended it to say why a zero dimension is a contradiction rather than an empty picture, which is the level of detail the entries around it carry.

PR #415, opened 2026-09-16
test: pin graphics Display and Debug renderings

### LINKED ISSUE #405: Six public items in the graphics module have no test, including every `DecodeError` message a user actually reads
**Today** — grepping the whole crate — unit tests, the integration suite
and doc examples — for these rendered strings finds them only in the
implementations themselves:

| item | what is unpinned |
| --- | --- |
| `Display for GraphicsProtocol` | `kitty` / `sixel` are never asserted |
| `Display for GraphicsFormat` | all five arms; the `f={value}` form has zero occurrences outside the impl |
| `GraphicsFormat::Other(u32)` | never constructed; no test sends an `f=` other than 24, 32 or 100 |
| `Display for DecodeError` | none of the five messages appears in any test |
| `Error for DecodeError` | never used |
| `Debug for Bitmap` | the deliberately terse `Bitmap WxH` form is unpinned |

The existing debug-rendering test asserts the payload's length and a few
substrings, not the protocol or format words. `Other(99)` does work today:
an `f=99` transmission yields it and renders as `f=99`.

**Why it is worth fixing** — these are the strings a failing test actually
reads. `DecodeError`'s wording is the entire point of
[#181](https://github.com/vyncint/termlens/issues/181), which asked that a
refusal name its reason, and it is load-bearing for whoever hits it and
completely unpinned. `Bitmap`'s `Debug` is deliberately short so that a CI
log stays readable, with nothing to stop a derive being restored over it.

**Fix** — one test in the graphics module's existing test module,
asserting the rendered text of each protocol, each format including
`Other(99)` driven through a real `f=99` transmission, each `DecodeError`
variant, and the `Bitmap` debug form.

**Done when**
- [ ] All five `DecodeError` messages and both protocol strings are
      asserted.
- [ ] `f=99` is exercised end to end, pinning the `Other` variant and its
      rendering.
- [ ] The `Bitmap` debug form is pinned.

---

Julia-b-work:
## What & why

Adds tests pinning the `Display` and `Debug` renderings in the graphics module
that had no coverage: the `GraphicsProtocol` and `GraphicsFormat` display
strings (including the `f={value}` arm), an unknown `f=99` format parsed
end-to-end through the builder, all five `DecodeError` messages plus its
`std::error::Error` impl, and the deliberately terse `Bitmap WxH` debug form.

Fixes #405

## Checklist

- [ ] Snapshot changes (if any) were reviewed with `cargo insta review`


vyncint (approved):
Thanks! Merging as is — nothing to change.

This is exactly what #405 asked for and a little more. Three things I liked:

Driving `f=99` through the real builder rather than constructing `GraphicsFormat::Other(99)` by hand is the right call, and you said why in the comment. A hand-built value would have pinned the `Display` arm while leaving a broken `f=` parse invisible; yours fails in both cases.

`fn assert_error<E: std::error::Error>() {}` pins the `Error` impl at compile time without a runtime assertion that could not really fail. Neat.

The comment on the `Bitmap` debug test says what the test is defending against — a `#[derive(Debug)]` restored over the manual impl, which would dump pixels into a CI log. That is the part a future reader needs and the part most such tests leave out.

I checked the tests bite: changing the `kitty` `Display` arm to `KITTY` fails `protocols_and_formats_render_as_their_names` with the expected left/right. They also build and pass under `--no-default-features` and under `--features decode` alone, which matters since `protocols_and_formats_render_as_their_names` is not feature-gated.

PR #416, opened 2026-09-16
docs(skill): add write failure triage

### LINKED ISSUE #399: The skill's failure-triage table has no row for the error every input method returns when the application has exited or stopped reading
**Today** — the triage table in `skills/termlens/SKILL.md` maps eleven
error first-lines to causes. `Error::Write` is not one of them:

```console
$ grep -c 'failed to send' skills/termlens/SKILL.md
0
$ grep -n 'failed to send' crates/termlens/src/error.rs
132:    #[error("failed to send {what}\n--- screen at the failed write ---\n{screen}")]
```

It has two causes with different fixes, and both are common:

```
failed to send Char('x') to `…/emit hi --exit 7` (the child is gone (exit code 7) and the terminal is closed)
failed to send a resize to `…/emit hi --exit 7` (the child is gone (exit code 7) and the terminal is closed)
failed to send … (the application is not reading its input, and the PTY buffer is full — no progress in 5s)
```

**Why it is worth fixing** — the skill is what an agent reads to triage a
red CI job, and this is the error a test hits whenever it types after the
application quit. It is easy to misread: "failed to send" sounds like a
harness fault rather than "your application already exited". The snippet
gate only compiles `rust` blocks, so nothing notices the table is short.

**Fix** — add two rows to that table in `skills/termlens/SKILL.md`: the
child-is-gone case, whose answer is to check `wait_exit()` and the screen
embedded in the error; and the not-reading case, whose answer is to wait
for the application to draw before typing. Quote the first lines exactly as
`error.rs` and the write path produce them — the table's value is that its
strings match byte for byte.

**Done when**
- [ ] Both reasons appear with their real first-line text.
- [ ] The wording matches the source exactly.
- [ ] `.github/scripts/check-skill-snippets.sh` still passes.

---

Rayan-and-beyond:
## What & why

Closes #399.

Adds the two missing `Error::Write` cases to the skill's failure-triage table: input attempted after the child has exited, and input blocked because the application stopped reading and the PTY buffer filled. The first-line examples mirror the current error/write-path wording, and the remedies point readers to `wait_exit()`/the embedded screen or to waiting until the application is ready for input.

## Validation

- `.github/scripts/check-skill-snippets.sh` — all 5 Rust blocks compile
- `git diff HEAD^ --check`

## Checklist

- [ ] Tests added/updated for the change


vyncint (approved):
Thanks! Merging as is.

I checked both strings against the source byte for byte, since that is the entire value of this table, and they match — including the two details most likely to be wrong in a hand-written row.

`exit code 7` is right. It comes from `ExitStatus`'s own `Display`, not from Rust's `std::process::ExitStatus`, which would have rendered `exit status: 7`. And `no progress in 5s` is right: that is a `Duration` Debug of the builder default, not a hand-written "5 seconds".

The part I think is most valuable is that you split one error string into two rows with two different fixes. A dead child sends you to `wait_exit()` and the screen the error already carries; a wedged one sends you to wait for the draw before typing. That distinction is the thing the issue was about, and a single row would have lost it.

Pointing the remedy column at `--- screen at the failed write ---` is a nice touch — it tells the reader the evidence is already in front of them.

PR #417, opened 2026-09-16
fix(input): allow wide UTF-8 mouse coordinates

### LINKED ISSUE #395: The 222-coordinate cap is applied to the UTF-8 mouse encoding too, so mode 1005 cannot reach the columns it exists for
**Today** — `CSI ?1005h` selects the UTF-8 mouse encoding precisely so that
coordinates above 222 can be sent: it writes `32 + 1 + coord` as a UTF-8
scalar rather than a bare byte. termlens encodes it correctly and then
refuses to use it, with a message naming an encoding the application did
not select. Against a 300-column terminal whose program sent
`CSI ?1000h CSI ?1005h`:

```console
click(222,1) OK
click(223,1) ERR: input not receivable: (223, 1) is unrepresentable in the legacy mouse encoding the application selected (max 222)
click(250,1) ERR: input not receivable: (250, 1) is unrepresentable in the legacy mouse encoding the application selected (max 222)
```

The click at 222 puts `c3 bf` on the wire — a correct two-byte UTF-8
scalar, not a bare `0xff` — so the encoder is already past the legacy
ceiling. The guard simply runs before the encoding is consulted: only the
SGR form returns early.

**Why it is worth fixing** — the crate's own documentation says the legacy
form is the one unusable past 222, and that UTF-8 exists to avoid exactly
that. A test driving a wide terminal — a 240-column runner, a split-pane
layout — cannot click the right half of the screen against an application
that chose 1005, and the diagnostic sends the reader to inspect the wrong
encoding.

**Fix** — in `crates/termlens/src/terminal.rs`, gate the 222 check on the
legacy encoding, and word the message so it names legacy only when legacy
is what was selected.

**Done when**
- [ ] A click past 222 under `?1000h` + `?1005h` succeeds and puts the
      two-byte UTF-8 coordinate on the wire.
- [ ] The legacy path still refuses 223 with its message.
- [ ] The existing UTF-8 mouse test in `crates/termlens/tests/input.rs`,
      which only clicks column 100, gains a case past 222.

---

Rayan-and-beyond:
## What & why

Closes #395.

The 222-coordinate ceiling belongs only to the legacy mouse encoding. UTF-8 mouse mode 1005 already encodes larger coordinates correctly, so this gates the limit on `MouseEncoding::Legacy` and adds a wide-terminal regression at column 250 while preserving the legacy refusal at column 223.

## Validation

- Red proof: the new column-250 UTF-8 regression failed with the legacy-limit error before the fix
- `cargo test -p termlens --test input utf8_mouse_reaches_past_the_legacy_limit -- --exact --nocapture`
- `cargo test -p termlens --test input mouse_reports_follow_the_utf8_encoding -- --exact --nocapture`
- focused test-target Clippy with `-D warnings`
- `git diff HEAD^ --check`

## Checklist


vyncint (approved):
Thanks! Merging.

Pinning the actual bytes rather than "no error" is what makes this test worth having — it proves the encoder produces a correct two-byte UTF-8 scalar, not merely that the guard stopped refusing. And keeping the legacy 223 refusal in the same test means the fix cannot be over-applied later by someone who only reads the happy path.

Writing a new dedicated test instead of extending the existing one, as the issue worded it, is the better shape and I would have asked for it the other way round.

I reverted the guard and it failed with the old diagnostic, naming the legacy encoding for a client that had selected 1005. The SGR path is untouched because it returns before the guard, and the reworded message can now only fire on the legacy encoding by construction, since those are the only three variants.

I pushed `89c1bb0` rewriting the CHANGELOG entry. It was one 84-column line saying what changed; every other entry in that file explains the consequence and wraps at about 74. The new text says why `CSI ?1005h` exists and what the refusal cost.

The batch including this passed `stress.yml`, 15 legs green — worth having, since this adds two PTY spawns.

PR #418, opened 2026-09-16
fix(emu): classify XTVERSION queries

### LINKED ISSUE #403: XTVERSION is the one query termlens neither answers nor names, so it is the only probe that can hang a test unexplained
**Today** — every unanswerable question is supposed to be classified, so
that a blocked application is named in the next timeout.
`CSI > q` (XTVERSION) is not in the classification table, so it becomes
nothing at all:

```
CSI > q   -> no reply; timeout says "timed out after 700ms while waiting for the
             screen predicate to hold"
CSI ? u   -> no reply; timeout says "… — note: the application queried the terminal
             (^[[?u) and received no answer; if it is blocked waiting for that
             reply, this is the cause"
```

Thirteen other unanswerable shapes were checked and every one produces the
note. The escape is consumed rather than printed, so nothing on the screen
hints at it either.

```console
$ grep -n "b'>'" crates/termlens/src/emu/seq.rs
1220:  (b'>', b'c') if params_empty || single(0) => Some(Query::SecondaryDa),
```

There is an arm for secondary device attributes and none for `q`.

**Why it is worth fixing** — XTVERSION is a standard startup probe, sent by
Neovim among others. `docs/LIMITATIONS.md` states the policy plainly: an
application blocked on an unanswered question is named in the next timeout
rather than left to hang unexplained. For this one probe that does not
hold, and it is the single hardest failure to diagnose because there is no
evidence anywhere.

**Fix** — one arm in `crates/termlens/src/emu/seq.rs`, classifying
`CSI > q` as an unanswerable query so it reaches the timeout note. It sits
safely beside the existing cursor-shape handling, which only claims `q`
when there is no prefix.

Actually *answering* it is a bigger decision — a new outbound claim, which
`docs/STABILITY.md` freezes — so classification alone is the change asked
for here.

**Done when**
- [ ] `CSI > q` and `CSI > 0 q` are classified as unanswerable queries and
      appear in the query-table unit test in `seq.rs`.
- [ ] A case in `crates/termlens/tests/queries.rs` shows the shape in the
      timeout note, in the style of the existing DECRQSS case.

---

Rayan-and-beyond:
## What & why

Closes #403.

Classifies `CSI > q` / `CSI > 0 q` (XTVERSION) as an unanswered terminal query so timeout diagnostics name the probe instead of reporting a generic wait timeout. This deliberately stops at classification; it does not add an XTVERSION reply or change the terminal's advertised capabilities.

## Validation

- Red unit proof: `CSI > q` produced no query before the fix
- Red integration proof: the timeout omitted XTVERSION before the fix
- `cargo test -p termlens --lib emu::seq::tests::recognizes_unanswerable_questions_with_their_shape -- --exact --nocapture`
- `cargo test -p termlens --test queries decrqss_and_palette_queries_are_named -- --exact --nocapture`
- focused library Clippy with `-D warnings`
- `git diff HEAD^ --check`

## Checklist


vyncint (approved):
Thanks! Merging as is.

Two things I liked. You reused the DA2 arm's parameter guard verbatim, so the new arm reads as a sibling of the one above it rather than a special case, and you placed it above the catch-all so it cannot be shadowed.

And you checked the collision the issue warned about instead of taking it on trust. I verified it independently: `CSI 2 SP q` still sets the cursor shape and produces no query, and the structural reason is sound — the space-intermediate branch returns before the query table is consulted, so a `q` with an intermediate can never reach your arm. DA2 still answers; a child sending `CSI > c` reads back the reply and unblocks.

Reverting the arm shows what the issue was about better than the issue did: before the fix the timeout carries no query clause at all, so there is nothing in the failure to suggest the application was waiting on an answer. After it, the shape is named.

Leaving the question *unanswered* and only classifying it was the right scope call. Answering `CSI > q` would be a new outbound claim, which `docs/STABILITY.md` freezes, and that is a bigger decision than this issue.

I rebased for the CHANGELOG. The whole batch, this included, passed `stress.yml` — 15 legs green — which matters here because the new case joins a test whose deadline also covers its spawn.

PR #419, opened 2026-09-16
fix(parse): refuse a control character in a grid row

### LINKED ISSUE #376: `Screen::parse` folds a control character into the previous cell, so `render --svg` writes a document no XML parser will open
**Today** — `parse_row` asks `unicode_width` for each character's width and
treats `None` as zero. `unicode_width` returns `None` for control
characters, not `Some(0)`, so an `ESC`, a tab or a `DEL` is appended to the
preceding cell exactly as a combining mark would be:

```
Screen::parse("size: 8x1  cursor: 0,0\na\u{1b}[2Jb")  → Ok
cell(0,0).contents() == "a\u{1b}"
cell(0,1).contents() == "["
to_ansi()             == "a\u{1b}[2Jb   \n"
```

Both renderings then carry the raw byte. The SVG is invalid XML:

```console
$ termlens render --svg ctrl.snap > out.svg; echo $?
0
$ xmllint --noout out.svg
out.svg:4: parser error : PCDATA invalid Char value 27
```

and the ANSI rendering clears the screen of anyone who pastes it, which is
the one thing `to_ansi`'s rustdoc says it is for.

**Why it is worth fixing** — `Screen::parse`'s rustdoc promises that
anything it does not understand "is an `Error::Parse` naming the line". The
snapshot format never writes a control character, and no file in the frozen
corpus contains one, so refusing them rejects nothing the format promises —
while accepting them produces an attachment that will not open in the
reviewer's browser.

**Fix** — in `crates/termlens/src/screen/parse.rs`, distinguish `None` from
`Some(0)`: a character with no width is an `Error::Parse` naming the line
and the character; a genuine zero-width combining mark keeps joining its
cell as it does now.

**Done when**
- [ ] A grid row containing `\u{1b}`, `\u{1}` or `\t` returns
      `Error::Parse` naming the line.
- [ ] A combining mark still joins the preceding cell — pin both in one
      test beside the existing wide-character-plus-combining-mark case.
- [ ] `crates/termlens/tests/compat.rs` still passes over the whole corpus.

---

maskjelly:
Closes #376.

`parse_row` asked `unicode_width` for each character's width and treated `None` as zero — but `None` is a control character and `Some(0)` is a zero-width combining mark, so an `ESC`, a tab or a `DEL` was appended to the preceding cell exactly as a mark would be:

```
Screen::parse("size: 8x1  cursor: 0,0\na\u{1b}[2Jb")  → Ok
cell(0,0).contents() == "a\u{1b}"
```

The raw byte then reached every rendering: `render --svg` wrote a document `xmllint` refuses to open (`PCDATA invalid Char value 27`), and `to_ansi` cleared the screen of whoever pasted it — the one thing its rustdoc says it is for.

The snapshot format never writes a control character and no file in the frozen corpus holds one, so refusing them rejects nothing the format promises. A character with no width is now an `Error::Parse` naming the line and the character; a genuine zero-width combining mark keeps joining its cell (`Some(0)` is still `Some`).

**Test:** beside the wide-character-plus-combining-mark case, one test pins both halves — `\u{1b}`, `\u{1}` and `\t` each return `Error::Parse` containing the line number and the character's `Debug`, while `a\u{301}b` still parses with the mark joined to `a`.

**Evidence.** Failing-before: with the guard reverted the test dies on `called Result::unwrap_err() on an Ok value: Screen(size: 8x1  cursor: 0,0\nab)` — the `ESC` sits folded between `a` and `b`. Passing-after: `crates/termlens/src/screen/parse.rs` tests 10/10. End to end, `render --svg` on the issue's control-character file now exits 2 with `line 2: '\u{1b}' is a control character; the snapshot format never writes one` (was exit 0 plus invalid XML), while the corpus's `wide.txt` — which holds a combining mark and CJK — still pipes through `xmllint` cleanly.

**Not done:** only `Screen::parse` refuses controls; a `Screen` built through serde or mutated by consumer code can still hold one and reach the renderings. That is outside this issue's scope, and hardening the renderers is the place to do it if wanted.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (48 result blocks, none failed), `cargo test -p termlens --all-features --test compat` over the whole corpus (5 passed, 1 ignored release tool).


vyncint (approved):
Thanks! Merging.

`None` versus `Some(0)` is the entire bug in one sentence, and the code comment says *why* — `unicode_width` answers `None` for a control character and `Some(0)` for a combining mark — rather than restating the `let-else`. That is the comment the next reader needs.

The regression risk here was the combining mark, and you pinned it in the same test: `a\u{301}b` still parses to one cell plus `b`, and both existing combining-mark tests pass. I checked the frozen corpus actually exercises it — `wide.txt` in all three version directories contains U+0301 — so `tests/compat.rs` is load-bearing evidence rather than a formality.

I reverted your change and the test failed the way the issue describes, with the ESC folded away between `a` and `b`. End to end on your branch, `termlens render --svg` on a screen holding `a\x1b[2Jb` now exits 2 naming the line and the character, where main exits 0 and writes an SVG no XML parser will open.

You were also right to scope out what this does not fix: a `Screen` built through serde or mutated by a consumer can still carry a control into the renderers. That is a fair follow-up and I would rather have it written down than silently assumed.

PR #420, opened 2026-09-16
fix(screen): keep the blank rows above the content in logical_text

### LINKED ISSUE #377: `logical_text()` drops every blank row above the first row with content, so it is not `text()` with the wraps undone
**Today** — the "is this the first line?" test inside `logical_text` is
`out.is_empty()`, which is still true after a blank row has been appended.
So leading blank rows vanish while trailing ones are kept. Measured through
a real PTY:

```
20x5, printf '\n\nhello world\n'
  text()         = "\n\nhello world\n\n"
  logical_text() = "hello world\n\n"

10x4, printf '\nabcdefghijklm\n'      (row 1 wraps)
  text()         = "\nabcdefghij\nklm\n"
  logical_text() = "abcdefghijklm\n"
```

An all-blank screen gives `text() == "\n\n"` and `logical_text() == ""`.

**Why it is worth fixing** — the rustdoc sells this as `text()` with the
wraps undone, "blank rows below the content included as blank lines just as
there". A TUI that leaves its top row blank — a title gap, a margin — has
every line of `logical_text()` shifted up, so `logical_text().lines()`
does not line up with grid rows and the two accessors disagree about the
shape of one screen. The method exists for the assertion that spans a
wrap, and that assertion is usually written by line index.

**Fix** — in `crates/termlens/src/screen.rs`, track whether a row has been
emitted with a flag rather than inferring it from the buffer being empty.
One function.

**Done when**
- [ ] On a screen with no wrapped rows, `logical_text() == text()` for any
      content, leading blanks and all-blank screens included.
- [ ] A case in `crates/termlens/tests/wrap.rs` asserts that equality, plus
      one wrapped screen where the two differ only by the joined wrap.

---

maskjelly:
Closes #377.

`logical_text` inferred "is this the first logical line?" from `out.is_empty()`, which is still true after a blank row has been appended — so every blank row above the first row with content vanished while trailing ones were kept:

```
20x5, printf '\n\nhello world\n'
  text()         = "\n\nhello world\n\n"
  logical_text() = "hello world\n\n"        ← was
```

A TUI that leaves its top row blank has every logical line shifted up, so `logical_text().lines()` no longer lines up with grid rows, against a rustdoc that sells the method as `text()` with the wraps undone.

A `first` flag now records whether a logical line has been emitted, so the two accessors agree on the whole grid whenever nothing wraps — leading blanks and an all-blank screen included — and differ only by the joined wrap when something does. The rustdoc's "blank rows below the content" becomes "above and below", which is what it now is.

**Tests** in `tests/wrap.rs`, both driven through the emit fixture: one asserts `logical_text() == text()` for a leading-blank screen and for an all-blank screen; the other is the issue's 10x4 wrap reproduction and asserts the two differ by the joined wrap and nothing else.

**Evidence.** Failing-before (source reverted, tests kept): `left: "abcdefghijklm\n"` vs `right: "\nabcdefghijklm\n"` for the wrapped case and `left: "hello world\n\n"` vs `right: "\n\nhello world\n\n"` for the leading-blank one — the issue's exact measurements. Passing-after: the file is 4/4.

**One honest gap:** the all-blank assertion is reached only after the leading-blank equality passes, so its individual red state was not observed in isolation; the old code returns `""` against `text() == "\n\n\n\n"` by the same mechanism and the issue measured exactly that.

**Stress note:** both new tests spawn the emit fixture, so CONTRIBUTING §3's stress requirement applies to them. I cannot dispatch workflows from a fork — could you run `stress.yml` on this branch?


vyncint (approved):
Thanks! Merging.

Driving both tests through the emit fixture rather than a parsed screen is what the issue asked for and what makes the result trustworthy: a parsed screen would have proved nothing about how a real grid reaches `logical_text`. Reverting the two `out.is_empty()` checks fails both tests with the issue's own numbers.

The rustdoc edit — "blank rows *above and below* the content" — is the easily-forgotten half, and the code comment explains why `out.is_empty()` was the wrong question rather than just describing the new flag.

I checked the one gap you were honest about. You noted the all-blank assertion sits after the leading-blank one and so had never been observed failing on its own. I wrote a throwaway test containing only the all-blank case and ran it against the unfixed source: `text()` gave four newlines and `logical_text()` gave the empty string. So that half bites too.

`stress.yml` was the other thing you asked for, since this adds three PTY spawns and a fork cannot dispatch it. I ran it on a branch carrying this and the rest of the batch: 15 legs green across Ubuntu, macOS and Windows. Your timing shape is fine — the only deadline is the generous builder timeout and nothing short covers a spawn.

Rebased for the CHANGELOG.

PR #421, opened 2026-09-16
fix(terminal): drop a Recorder deregisters it

### LINKED ISSUE #396: A `Recorder` dropped without `stop()` keeps collecting frames for the life of the terminal
**Today** — `Recorder::stop` is the only place a recorder deregisters
itself, so an early return between `record()` and `stop()` leaves the
reader thread cloning a whole `Screen` into a buffer nobody can read, until
the `Terminal` is dropped. The only `impl Drop` in the crate is
`Terminal`'s. Two identical three-second runs, differing only by
`let r = t.record(); drop(r);`:

```
kept    repaints=138177   rss 4096 KB -> 5872 KB
dropped repaints=148666   rss 5184 KB -> 206416 KB
kept    repaints=140105   rss 4176 KB -> 6208 KB
dropped repaints=124483   rss 5632 KB -> 206576 KB
```

200 MB is the default record budget being filled by an unreachable object.

**Why it is worth fixing** — the idiomatic shape is
`let rec = t.record(); … t.wait_frame(…)?; let r = rec.stop()?;`, so any
`?` or panic before `stop()` triggers it. It is silent, and the test that
pays for it is whichever one runs longest afterwards. It also costs a
full-grid clone on every repaint, forever.

**Fix** — add `impl Drop for Recorder` in
`crates/termlens/src/terminal.rs`, doing the same deregistration `stop`
does. `stop` consumes `self` without destructuring it, so the
deregistration simply runs twice and the second is a no-op; `stop` itself
does not need to change. Say in the rustdoc that dropping a recorder stops
the recording.

**Done when**
- [ ] After `drop(t.record())`, a terminal driven through many frames grows
      no more than one with no recorder at all.
- [ ] A test in `crates/termlens/tests/record.rs` covers it — for example a
      dropped recorder, then frames, then a second `record().stop()` that
      returns only its own.
- [ ] `Recorder`'s rustdoc says what dropping it does.

---

maskjelly:
Closes #396.

`Recorder::stop` was the only place a recorder took itself out of the reader thread's delivery list, so the idiomatic shape — `let rec = t.record(); … t.wait_frame(…)?; rec.stop()?` — leaked the recorder on any `?` or panic before `stop`: the reader kept cloning a whole grid into a buffer nobody could read, filling the default 2M-cell record budget (about 200 MB measured) for the life of the terminal. The test that paid for it was whichever one ran longest afterwards.

`stop` and the new `Drop` now share one `deregister`, so a stopped recorder double-deregisters by construction and the second call is a no-op — `stop` itself needed no change to its error path and keeps its single critical section. `Recorder`'s rustdoc and `Terminal::record`'s now say that dropping a recorder stops the recording.

**The test is a memory bound, not an API shape.** `tests/record.rs` gains a counting global allocator, because what this bug costs is *retention*, not allocation — the reader clones a screen that shares its grid through an `Arc`, so a recorder that is never deregistered allocates nothing new and simply never lets frames go, which RSS measures and portable code does not. It drives 64 frames at 200x50 with no recorder (the control), then 64 more after `drop(t.record())`, and requires the dropped phase to grow no more than the control plus 8 MiB. It then starts a *later* `record()`, drives one frame and asserts `stop()` returns exactly that one, so the dropped recorder cannot shadow a live one.

**Evidence.** Without `Drop` the dropped phase grew 26,902,496 B against a 3,846,240 B control (23 MB over the bound, deterministically); with it the same phase grew 2,048 B. The assertion is `control + 8 MiB`, so the noise headroom is more than 3× either side. The new test ran 10× debug and 5× release at `--test-threads=16`, all green; it carries the same `windows` ignore as the other recording-content tests (#149, ConPTY closes the DEC 2026 bracket before its content).

**Stress note:** the new test spawns a PTY, so CONTRIBUTING §3's stress requirement applies; I cannot dispatch workflows from a fork — could you run `stress.yml` on this branch?

**Not done:** `mem::forget(rec)` still bypasses deregistration, as `forget` bypasses every `Drop`; noted rather than special-cased.


vyncint (approved):
Thanks! Merging — this is the best-engineered PR in the batch.

The diagnosis is the subtle part: what leaks is *retention*, not allocation, because the reader clones a `Screen` that shares its grid through an `Arc`. That is why the test measures live heap rather than allocation count, and it is the reason a naive peak-RSS check would have missed it. I confirmed that: `/usr/bin/time -l` cannot separate the two cases at the test's own scale, while your allocator counter reports 2 KB against 27 MB.

I reproduced the leak two independent ways — your metric, and an external RSS probe over 400 frames, which showed 103 MB of growth without the fix and 2.4 MB with it.

Factoring `stop` and `Drop` onto one `deregister` is the right shape; duplicating the `retain` would have let the two drift.

I looked hard at the `Drop` impl, since that is where this kind of fix goes wrong, and it is sound. It takes only the shared lock and never touches the recorder's own, so it cannot invert the reader thread's order. The lock swallows poisoning crate-wide, and the closure body has no panicking path, so it cannot double-panic during an unwind — I tested a recorder dropped mid-panic and it unwound cleanly with a later `record()` still working. Two simultaneous recorders with one dropped mid-stream also behaved: the survivor returned exactly its own frames.

One conscious "yes" from me: the test installs a global allocator for that whole test binary, so the counter is process-global and the other seven tests perturb it. The margin is 27 MB against an 8 MiB bound, and it survived ten runs at 16 threads plus the full `stress.yml` batch, 15 legs green. I am accepting it, but that is the one thing here that could flake under an unlucky schedule.

Your `mem::forget` caveat is correct and worth having written down. And your follow-up idea is a good one — `#[must_use]` on `Terminal::record` would now catch a bare `t.record();`, which since this change is a silent no-op rather than a leak. Worth a small issue if you want to file it.

Rebased for the CHANGELOG and wrapped one 79-column line.

PR #422, opened 2026-09-16
fix(diff): pad the frame by display width, not char count

### LINKED ISSUE #380: The diff rendering pads rows by character count, so a wide character knocks the right-hand column out of line
**Today** — the before and after columns are padded with `{before:<width$}`,
which Rust counts in `char`s, while the marker line below is ASCII and pads
to the same number of *display* columns. On an 8-column screen:

```console
$ # row is 東京ab, changed to 東京aZ
size: 8x1   cursor: 0,0
  0 │東京ab    │東京aZ
    │     ^  │     ^

$ # the same shape, all ASCII
size: 8x1   cursor: 0,0
  0 │wxyzab  │wxyzaZ
    │     ^  │     ^
```

In the first, the text line's closing `│` sits two display columns right of
the marker line's. The `^` markers themselves land correctly; only the
frame drifts.

**Why it is worth fixing** — `Screen::diff` is the readable-failure half of
the snapshot workflow, printed into CI logs and painted by `termlens diff`.
Wide characters are exactly what a PTY harness exists to get right, and
they are the case where the two columns stop lining up and become hard to
read side by side.

**Fix** — in `crates/termlens/src/screen/diff.rs`, pad by display width
using `unicode_width`, which is already a direct dependency, instead of the
`{:<width$}` formatter. One function.

**Done when**
- [ ] For a row of wide characters, the `│` separators on the text line and
      on the marker line sit at the same display column.
- [ ] A test asserts the rendered lines' display widths agree for a CJK row
      and an emoji row.
- [ ] The existing diff snapshots are reviewed with `cargo insta review`,
      not blind-accepted.

---

maskjelly:
Closes #380.

The row columns of `Screen::diff` were padded with `{:<width$}`, which counts `char`s, while the marker line below is ASCII and padded to the same number of display columns. `row_text` emits one `char` per wide glyph, so on a row holding k wide characters the text line's closing `│` sat k display columns right of the marker line's:

```
size: 8x1   cursor: 0,0
  0 │東京ab    │東京aZ       ← the closing │ is two columns right of the marker's
    │     ^  │     ^
```

Both frame sites now pad through one `pad()` helper using `unicode_width` (already a direct dependency), saturating rather than truncating. The `after` column and the trailing markers were already unpadded and stay that way — nothing follows them, and padding would only add trailing whitespace to every rendered line.

**Tests:** a unit test in `diff.rs` (which had no test module; the sibling modules do) renders a CJK row and an emoji row and asserts three things per case — the exact expected frame, that the display column of every `│` agrees between the text line and its marker line, and that the two lines' whole display widths agree.

**Evidence.** Failing-before: `the frame for "東京ab": left "…│東京ab    │…"` vs `right "…│東京ab  │…"` — the fix removes exactly the two extra char-pad spaces, and the emoji case one. Passing-after: the test is green and `cargo test -p termlens --test export` shows the existing diff snapshot untouched.

**Snapshots:** none changed. The checked-in diff snapshot is pure ASCII (`Counter: 1/2`, `> Quit`), where char count equals display width, so `pad()` produces byte-identical output. No `.snap.new` was produced by any run.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (48 result blocks, none failed).


vyncint (approved):
Thanks! Merging.

Asserting the *display column of every* `│` rather than eyeballing an expected string is what makes this test encode the bug instead of the symptom. Keeping the exact-frame assertion alongside it means a future change has to explain itself twice, which is the right amount of friction here.

I reverted the two `{:<width$}` uses and the test failed showing exactly the two extra pad spaces removed for the CJK row and one for the emoji row, with the marker line unaffected. No snapshot moved — I checked the committed diff snapshot is pure ASCII, where char count and display width agree, so the change is a no-op there by construction rather than by luck.

Noticing that the `after` column and the trailing markers were already unpadded, and deliberately leaving them that way because padding would only add trailing whitespace to every line, is the kind of restraint that keeps a fix from growing into a reformat.

I only had to rebase this onto main for the CHANGELOG, since several entries landed under `[Unreleased]` while this was open.

Issue #360, opened 2026-09-16
Tight per-test timeouts flake on the Windows leg under thread contention


vyncint:
**Today** — `stress.yml` turns up a recurring failure on the Windows leg
whose shape is always the same: a test with a **sub-second builder timeout**
times out with an **empty screen**, because the child had not written
anything yet. It is not the same test twice, which is what makes it a class
rather than a bug in any one of them.

Two measurements, on different branches and different thread counts:

| date | run | shard | tests | iteration |
| --- | --- | --- | --- | --- |
| 2026-09-10 | [34527017048](https://github.com/vyncint/termlens/actions/runs/34527017048) | `windows-latest, 8 threads` | `wait_idle_for_overrides_the_builder_default` | — |
| 2026-09-16 | [35101695488](https://github.com/vyncint/termlens/actions/runs/35101695488) | `windows-latest, 16 threads` | `unanswerable_queries_turn_timeouts_into_diagnoses`, `the_responder_can_be_disabled_and_says_what_went_unanswered`, `all_unanswered_queries_are_named`, `wait_frame_timeouts_carry_the_query_note` | 21/25 |

The 2026-09-16 diagnostic, which is the whole story in four lines:

```
query not named in: timed out after 500ms while waiting for the screen predicate to hold
--- screen at timeout ---
size: 80x24  cursor: 0,0
                          ← nothing on the grid at all
```

The four `queries.rs` tests assert that a timeout **names the query that
went unanswered**. They cannot: the child never emitted the `--csi 14t` or
`--csi 6n` within the 500ms the builder was given, so there was no query to
name. The assertion is about the diagnosis; what failed is process startup.

**Why it is worth fixing** — this is the failure mode that teaches people to
ignore a red stress run, and `stress.yml` is the instrument this repository
relies on for exactly the bugs it is best at finding. A shard that goes red
for contention rather than for a defect costs a re-run every time and
eventually costs the run its credibility.

It is also the *only* remaining stress failure not attributable to a test
added on 2026-09-16 (those two are fixed in #359, one of them proven
deterministically).

**Fix** — the honest options, roughly in order of preference:

1. Give the Windows leg a **spawn-latency floor**: the 500ms in these tests
   is chosen to keep the suite fast, and ConPTY's process startup under
   16-way contention plainly does not fit in it. A platform-aware minimum
   (`cfg!(windows)` → a larger floor) keeps the test's meaning — the
   timeout still has to name the query — while not asserting a schedule.
[...]

PR #423, opened 2026-09-16
ci(stress): run the whole suite under all features

### LINKED ISSUE #382: The stress workflow runs the default feature set, so the one `regex` wait and the eighteen `decode` tests are never stressed
**Today** — `stress.yml` describes itself as the whole test suite, 100
times, and CONTRIBUTING §3 makes passing it the condition for touching wait
semantics. It builds and runs without `--all-features`, while the `test`
job in `ci.yml` uses them:

```console
$ grep -n 'cargo build\|cargo test' .github/workflows/stress.yml
134:  run: cargo build --workspace --release --all-targets
148:  cargo test --workspace --release -- --test-threads="$THREADS" …
$ grep -n 'cargo test --workspace --all-features' .github/workflows/ci.yml | head -1
67:  - run: cargo test --workspace --all-features
```

The two selections differ by 22 tests: the eighteen `decode` tests, three
`regex` ones and a doctest. One of them is a wait test —
`patterns::wait_until_matches_returns_the_screen_it_matched` in
`crates/termlens/tests/search.rs` — which drives a child that sleeps and
then asserts on a deadline. That is exactly the class CONTRIBUTING §3 and
[#360](https://github.com/vyncint/termlens/issues/360) are about, and it
has never been stressed.

**Why it is worth fixing** — `wait_until_matches` is a promised public wait
in `docs/STABILITY.md`. The gate meant to clear wait code before a tag does
not compile it, so the workflow's own summary line overstates what it
proved. This is also the cheapest possible fix for the largest gap.

**Fix** — add `--all-features` to both steps in
`.github/workflows/stress.yml`, and correct the header comment so it says
which suite is stressed.

**Done when**
- [ ] Both the build and the test step carry `--all-features`.
- [ ] `cargo test --workspace --release --all-features -- --list | grep -c ': test$'`
      matches what the `test` job runs.
- [ ] A dispatch run on a branch is green before the PR is merged, since
      this is the first time those 22 tests see the loop.

---

maskjelly:
Closes #382.

`stress.yml` calls itself the whole test suite, 100 times, and CONTRIBUTING §3 makes passing it the condition for touching wait semantics — but both steps ran without `--all-features`, while ci.yml's `test` job uses them. The two selections differ by 22 tests: the eighteen `decode` tests, three `regex` ones and a doctest — including `patterns::wait_until_matches_returns_the_screen_it_matched`, the one wait test that sleeps and then asserts on a deadline, exactly the class this workflow exists to clear, and it had never been stressed.

Both the build and the test step now carry `--all-features`, and the header names the suite it stresses, so a green run stops overstating what it proved.

**Evidence:** the selections were counted both ways — 520 tests by default, 542 with `--all-features` — and the 542-name list is byte-identical to what the debug `test` job runs. A single release pass (`cargo test --workspace --release --all-features -- --test-threads=4`) is green locally, with the previously-unstressed `wait_until_matches…` and the `decode` tests included.

**The dispatch-run Done-when needs you:** I cannot dispatch workflows from a fork, and this is the first time those 22 tests enter the loop, so `stress.yml` on this branch is the check that matters. Could you dispatch it? (If enabling Actions on my fork and dispatching there is acceptable to you, that works too — say which you prefer.)

**Gates:** the release build with `--all-features` and a release test pass were run locally; `check-ci-gates-listed.sh` still passes.


vyncint (approved):
Thanks! Merging — and the third box is now ticked.

I pushed your branch to the base repository and dispatched `stress.yml` on it, since a fork cannot dispatch. **All fifteen legs green**: Ubuntu, macOS and Windows across 1, 2, 4, 8 and 16 test threads, 100 iterations each. That matters more than usual here, because those 22 tests had never been through the loop before — including `patterns::wait_until_matches_returns_the_screen_it_matched`, which sleeps in the child and asserts on a deadline, exactly the shape #360 was about.

Your count checks out precisely: 520 tests by default, 542 with all features, and the 22-test delta is 18 `decode`, 3 `regex` and one doctest. You verified it both ways and listed the names rather than asserting the number, which is why I could confirm it in one command instead of re-deriving it.

The header comment is the part that will keep this fixed. It now says *why* `--all-features` is what makes "the whole suite" true, so the next person cannot quietly re-narrow it.

Issue #341, opened 2026-09-10
Every relative link in the README is a 404 on crates.io


vyncint:
**Today** — `crates/termlens/Cargo.toml` sets `readme = "../../README.md"`, so the root README is shipped as this crate's readme. crates.io rewrites a *relative* link against the **crate's directory** in the repository rather than the repository root, so

```
[docs/DESIGN.md](docs/DESIGN.md)
```

is served on the crate page as

```
https://github.com/vyncint/termlens/blob/HEAD/crates/termlens/docs/DESIGN.md   → 404
```

There is no `crates/termlens/docs/`. **All ten relative links on the published 0.10.2 page are dead**, checked one at a time against the live rendering (`/api/v1/crates/termlens/0.10.2/readme`):

| target as published | |
|---|---|
| `crates/termlens/docs/DESIGN.md` | 404 |
| `crates/termlens/docs/STABILITY.md` | 404 |
| `crates/termlens/docs/LIMITATIONS.md` | 404 |
| `crates/termlens/docs/BACKENDS.md` | 404 |
| `crates/termlens/skills/termlens/SKILL.md` | 404 |
| `crates/termlens/CHANGELOG.md` | 404 |
| `crates/termlens/CONTRIBUTING.md` | 404 |
| `crates/termlens/SECURITY.md` | 404 |
| `crates/termlens/LICENSE-MIT`, `LICENSE-APACHE` | 404 |
| `crates/termlens/.github/workflows/stress.yml` | 404 |

**Nothing here could see it.** The same file renders correctly on GitHub, where it really is at the root, so the only broken surface is the one most new readers arrive at — and it has been broken for every release that shipped this README.

`termlens-cli` is unaffected: it has its own `crates/termlens-cli/README.md` with no relative links. The crate's own rustdoc is unaffected too — every link in the doc comments is an intra-doc link (`[`Terminal`](crate::Terminal)`), not a file path.

## Fix

Absolute links, since the same file is rendered from two different roots and no relative form is correct in both. Plus `.github/scripts/check-readme-links.sh` in `ci.yml`, which does two things:

- refuses a relative link, naming the absolute URL to use instead;
- asserts every absolute target **still exists in the checkout** — offline and deterministic, so it also catches a renamed or deleted file, which a checker that only asks GitHub would not.

## Scope

**The fix only reaches future releases.** crates.io renders each version's readme as it was published, so the 0.10.2 page and every earlier one stay broken. Nothing can change them.

## The same bug is in launchbound

[...]

PR #424, opened 2026-09-16
build: exclude the test suite from the published crate

### LINKED ISSUE #385: The published crate ships 86 test files it cannot run, because the fixture binaries they spawn are not in the package
**Today** — `crates/termlens/Cargo.toml` sets neither `include` nor
`exclude`, so `cargo package` takes the whole directory:

```console
$ cargo package --list -p termlens | wc -l
109
$ cargo package --list -p termlens | grep -c '^tests/'
86
$ cargo package --list -p termlens | grep -c '^tests/compat/'
37
```

Most of the tarball is therefore a test suite that cannot run from it. The
integration tests spawn fixture binaries through a workspace-sibling
`cargo build -p <fixture>`, and `fixtures/` is not in the package — so
inside the unpacked crate that build fails with "package ID specification
did not match any packages". Thirty-one of the test files use those
fixtures. The 37 files under `tests/compat/` are the frozen
release-engineering corpus, which is evidence about *this* repository.

**Why it is worth fixing** — most of every download, and of what a distro
packager vendors, is dead weight nobody can execute. `docs/STABILITY.md`'s
claim that `tests/compat.rs` checks the format is about the repository, not
the tarball, so shipping the corpus buys a consumer nothing.

**Fix** — add `exclude = ["tests/"]` to `crates/termlens/Cargo.toml`, with
a comment recording why: the tests need workspace-sibling fixture binaries
that the package cannot carry. Keep `examples/inspect.rs` and `README.md`.
Check `cargo package -p termlens` still verifies, since that step builds
the packaged crate.

**Done when**
- [ ] `cargo package --list -p termlens` shows no `tests/` entries.
- [ ] `cargo package -p termlens` verifies, and the doctests still run.
- [ ] The manifest says why in a comment, so the next person does not
      "restore" them.

---

maskjelly:
Closes #385.

The package sets neither `include` nor `exclude`, so 86 of its 109 files were the integration suite — which cannot run from the tarball at all: the tests spawn fixture binaries built from workspace siblings that `publish = false` keeps out of the package, so inside the unpacked crate that build fails with "package ID specification did not match any packages". The 37 files under `tests/compat/` are frozen release-engineering evidence about *this* repository, not the crate.

An `exclude = ["tests/"]` leaves 23 files — `src`, `examples/inspect.rs`, `README.md`, and the manifests — and the manifest comment records why, so the tests are not "restored" to the tarball by the next person who reads this issue upside down.

**Evidence:** `cargo package --list -p termlens` went from 109 entries (86 under `tests/`, 37 under `tests/compat/`) to 23 with none; `examples/inspect.rs` and the readme entries are still there. `cargo package` verifies (the packaged crate compiles, 726.6 KiB packed from 207.9 KiB compressed) and the doctests — which come from `src/`, not `tests/` — still run (39 passed, 2 ignored). The full suite is untouched.

**Two honest notes:** the package step needs the crates.io index over the network to verify (it succeeded here), and `cargo package` now prints one `warning: ignoring test tests/x.rs as it is not included in the published package` per excluded file — inherent to `exclude` plus test auto-discovery. Silencing those would mean `autotests = false` with explicit `[[test]]` sections, which is a much larger change I left alone; say the word if you want it.


vyncint (approved):
Thanks! Merging.

I verified the part most likely to break, which is `cargo package` itself — it builds the packaged crate, so an over-broad exclude shows up there rather than in the file list. It verifies, the tarball is 23 files with nothing under `tests/`, and `README.md` and `examples/inspect.rs` both survive. Doctests are unaffected at 39 passed, 2 ignored.

The manifest comment is the part I would have asked for if it were missing. It records *why* the tests cannot run from the tarball — the fixture binaries are workspace siblings a package cannot carry — rather than just recording the decision. That is what stops the next person restoring them in good faith.

Your two honest notes in the PR body were both accurate: the packaging step needs the network, and `cargo package` now prints 33 "ignoring test target" warnings. Those are cargo warnings, not rustc, so `cargo publish --locked` will print them and still succeed. I agree with leaving `autotests` alone.

I pushed one commit, `8af4fdd`, adding the `[Unreleased]` entry. CONTRIBUTING §7 asks for one on any user-facing change, and the published tarball going from 109 files to 23 is one — it is what every `cargo add termlens` and every distro packager downloads. #341's packaging fix carried an entry for the same reason. I filed it under **Changed** rather than **Fixed**, since nothing was broken, the contents moved.

PR #425, opened 2026-09-16
fix(terminal): a resize to the current size is a no-op

### LINKED ISSUE #397: Resizing to the size the terminal already has advances the frame cursor although no `SIGWINCH` is sent, so the `wait_frame` the docs recommend can never succeed
**Today** — `resize` advances the frame cursor unconditionally, so that a
frame drawn at the old size cannot answer the resize. But `TIOCSWINSZ`
raises `SIGWINCH` only when a dimension actually changes, on both Linux and
macOS. Resize a 60x12 application to 60x12 and the application never hears
anything, while the cursor has moved past its last frame:

```
same-size resize      -> timed out after 1s while waiting for a complete frame
                         matching the predicate — the application has not completed
                         a repaint since the frame this terminal last returned
                         (1 complete frame in total)…
one-column difference -> frame arrived, size (61, 12)
```

**Why it is worth fixing** — `resize`'s own rustdoc recommends `wait_frame`
here and calls it "unconditionally safe", because a resize advances the
cursor. For a no-op resize that reasoning inverts: the advance is what
makes the wait impossible. The no-op case is what a parameterised or
loop-driven layout test produces — a list of sizes with a repeat in it, or
a resize back to the builder's size — and the failure is a full-deadline
hang whose message blames the predicate and the application. Every existing
resize test changes a dimension, so nothing catches it.

**Fix** — in `crates/termlens/src/terminal.rs`, compare the requested size
against the current grid inside the same critical section and return
`Ok(())` without touching the ioctl or the frame cursor when they are
equal. Adjust the two sentences of rustdoc that describe the cursor
advance, so a no-op resize is documented as a no-op.

**Done when**
- [ ] A resize to the current size leaves the frame cursor where it was,
      and a following `wait_frame` still sees the frame the application
      last drew.
- [ ] A resize that does change a dimension still advances the cursor, and
      the existing frames tests pass unchanged.
- [ ] A new case in `crates/termlens/tests/frames.rs` pins the same-size
      behaviour.

---

maskjelly:
Closes #397.

`resize` advanced the frame cursor unconditionally, so that a frame drawn at the old size could not answer the resize. But `TIOCSWINSZ` raises `SIGWINCH` only when a dimension actually changes, so resizing a 60x12 terminal to 60x12 advanced the cursor past the last frame while the application was never told to repaint: the `wait_frame` the rustdoc recommends — and called "unconditionally safe" — could then only run out its deadline, blaming the predicate and the application for a no-op. That is exactly the shape a parameterised layout test produces when its size list repeats one, and every existing resize test changes a dimension, so nothing caught it.

The requested size is now compared against the grid inside the same critical section that owns the cursor; on equality the closure returns early and neither the ioctl, nor the generation touch, nor the cursor move happens. `check_size` and the child-alive check still run first, so a same-size resize after the child exits is still `Error::Write` — the documented refusal is unchanged.

**Docs:** the rustdocs on `resize` and `wait_frame` and the DESIGN.md wait contract say which resize advances the cursor and which is a no-op; "unconditionally safe" is gone, replaced by the two-branch statement.

**Test:** a same-size resize followed by `wait_frame` for the frame the application last drew must return promptly — with a generous bound against the deadline, and the frame is provably published and unconsumed first by waiting on the repaint count, not on text. Without the fix it times out with the issue's own message (`the application has not completed a repaint since the frame this terminal last returned (1 complete frame in total)`); with it, the frame comes back. The new test ran 10× plus its file 5× more, all green; the positive half — a dimension-changing resize still advances the cursor — is pinned by the existing frames tests, unchanged.

**Stress note:** this touches frame-cursor semantics and the new test spawns a PTY, so CONTRIBUTING §3 applies; I cannot dispatch workflows from a fork — could you run `stress.yml` on this branch?

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (538 tests, 0 failed), `RUSTDOCFLAGS="-D warnings" cargo doc --workspace --all-features --no-deps` — all green.


vyncint (approved):
Thanks! Merging as is — nothing needed changing beyond a rebase.

Two details make this better than the issue asked for.

Gating the precondition on `s.repaints() == 1` rather than on the screen text is what makes the test prove the right thing: it guarantees the frame is published *and unconsumed* when the resize lands, so an unwanted cursor advance would have something to hide behind. A text-based precondition would have passed either way.

And putting the size comparison inside the same critical section that owns the frame cursor, rather than reading the size before taking the lock, is exactly the discipline the surrounding comment in `resize` was built around.

Reverting the change fails the new test after the full deadline with the issue's own message verbatim, and both existing resize tests still pass, so a real size change still advances the cursor.

The `docs/DESIGN.md` edit is accurate and the sweep is complete — `grep -rn "unconditionally safe"` finds nothing left anywhere in the repo, and both rustdocs were updated rather than just one. That is the half most people forget.

I also checked something the PR does not mention: the early return skips the ioctl, which carries pixel dimensions. Those are a pure function of the cell count and the builder-only cell size, so an unchanged size cannot imply changed pixels. Skipping it loses nothing.

`stress.yml` was the right thing to ask for, since this changes frame-cursor semantics. I ran it on a branch carrying this and the rest of the batch: 15 legs green.

PR #426, opened 2026-09-16
fix(emu): preserve parameterless unsupported sequences

### LINKED ISSUE #394: An unhandled sequence with no parameters is recorded with a parameter the application never wrote — `^[[s` is listed as `^[[0s`
**Today** — the underlying parser always presents a zeroed accumulator, so
a sequence written with no parameters is recorded as though it had one:

```
sent ^[[s    -> unsupported ["^[[0s"]      (SCOSC, very common in shell prompts)
sent ^[[>q   -> unsupported ["^[[>0q"]     (XTVERSION)
```

**Why it is worth fixing** — the module's header and its own test name both
say the record is the sequence "in its written form", and these strings are
what a user greps for or pastes into xterm's control-sequence reference.
`^[[>0q` matches nothing there. `CSI s` and `CSI u` really are
unimplemented, so this diagnostic is the one a user reaches for to find
that out, and it is rendered wrong.

**Fix** — in `crates/termlens/src/emu/unhandled.rs`, render no parameter
text when the parameter list is a single group holding only a zero. An
omitted parameter and an explicit `0` mean the same thing to every
terminal, so collapsing them to the written form is safe.

**Done when**
- [ ] `^[[s`, `^[[>q` and `^[[?h` are recorded in exactly that form.
- [ ] `^[[20h`, `^[[?69h` and `^[[58;5;1m` are unchanged, and the existing
      unit tests plus `crates/termlens/tests/unsupported.rs` pass.
- [ ] A unit test feeds a parameterless sequence and asserts the shape.

---

Rayan-and-beyond:
## What & why

Closes #394.

`vt100` presents an omitted CSI parameter as a single zeroed accumulator, so the unsupported-sequence recorder was inventing a `0` that the application never wrote. Treat that one synthetic/default-zero shape as parameterless when rendering the diagnostic, while leaving real parameter lists unchanged.

The regression covers bare, private, and secondary parameterless CSI forms alongside real single/private/multi-parameter sequences.

## Validation

- Red proof: the new regression produced `^[[0s`, `^[[>0q`, and `^[[?0h` before the fix
- `cargo test -p termlens --lib emu::unhandled::tests -- --nocapture` — 5/5 passed
- `cargo test -p termlens --test unsupported -- --nocapture` — 7/7 passed
- `git diff HEAD^ --check`

I did not run the full workspace Clippy matrix locally because the shared development VM was at its disk-headroom limit; that checklist item is intentionally left unchecked.

## Checklist


vyncint (approved):
Thanks! Merging.

The inline comment gives the *reason* — vte pushes a zeroed accumulator before dispatch, so an omitted parameter and an explicit `0` arrive identically — rather than restating the code. That is the house style, and it is the fact the next reader needs, because the obvious question about this change is "are you sure you are not erasing a real zero?" The answer is that the parser cannot tell, and your comment says so.

I checked that claim rather than taking it on trust, and it holds: `^[[s` and `^[[0s` both arrive as a single zero parameter, as do `^[[>q`/`^[[>0q` and `^[[?h`/`^[[?0h`. Multi-parameter forms keep their zeros, so `^[[0;0s` is untouched. Reverting your guard fails the test with the whole before/after list, which is a good diagnostic.

Putting the must-change and must-not-change cases in one assertion means a regression in either direction fails the same test.

I rebased onto main for the CHANGELOG and wrapped its first line, which was 80 columns against the file's ~74.

PR #427, opened 2026-09-16
fix(emu): report unsupported title stack operations

### LINKED ISSUE #393: The xterm title stack is silently ignored, so `title()` reports the title a program pushed away
**Today** — the unhandled-sequence filter treats every `CSI … t` as
implemented, whatever its first parameter. The responder answers only the
size reports, so the title stack is dropped with no trace:

```
sent: ^[]2;First BEL  ^[[22;0t  ^[]2;Second BEL  ^[[23;0t  A
title()     = "Second"
unsupported = []
```

On xterm, `CSI 22;0t` pushes the title and `CSI 23;0t` pops it, so after
that sequence the title is `First`. termlens reports `Second` and says
nothing was unsupported.

**Why it is worth fixing** — bracketing a run with push and pop is exactly
what a well-behaved TUI does so the user's terminal title survives it, so
this is not an exotic sequence. A test asserting on `Screen::title()`
around such a program gets a stale answer *and* an empty `unsupported()`,
which is the combination the accessor exists to prevent.

**Fix** — in `crates/termlens/src/emu/unhandled.rs`, split the `t` final
byte out of the blanket exemption and gate it on the first parameter being
one the responder actually handles. Implementing the stack itself is a
larger change and not required here: being listed as unsupported is the
honest answer, and is what this issue asks for. If you would rather
implement it, say so in the PR and keep the two changes in separate
commits.

**Done when**
- [ ] `^[[22;0t` and `^[[23;0t` appear in `unsupported()`.
- [ ] `^[[18t` and `^[[14t` still do not, and `crates/termlens/tests/queries.rs`
      passes unchanged.
- [ ] A test pins the title-stack case.

---

Rayan-and-beyond:
## What & why

Closes #393.

The unhandled-sequence filter treated every plain `CSI … t` sequence as implemented even though the responder only handles the 14, 16, and 18 size reports. Restrict that exemption to those three reports so xterm title-stack operations such as `CSI 22;0 t` and `CSI 23;0 t` are visible through `Screen::unsupported()`.

A focused regression pins both title-stack operations and keeps the three supported size queries out of the unsupported record.

## Validation

- Red proof: the new regression returned an empty unsupported list before the fix
- `cargo test -p termlens --lib emu::unhandled::tests -- --nocapture` — 5/5 passed
- `cargo test -p termlens --test queries -- --nocapture` — 23/23 passed
- `git diff --check`

Full workspace Clippy was not run locally because the shared development VM remains at its disk-headroom limit.

## Checklist


vyncint (dismissed):
Thanks! Merging.

You found the right seam. Splitting `t` out of the blanket exemption, rather than special-casing 22 and 23, is what makes this a rule instead of a patch — and the test pins both directions in one function, so a regression either way fails the same assertion.

I pushed `c5f4fad`, because the allowlist was one step too narrow. It kept only the three size reports the responder *answers* (14, 16, 18), but `seq.rs` also classifies `CSI 11/13/19/20/21 t` as unanswerable queries, which the timeout note names. This module's header says the filter exists so the record excludes every query the responder answers **or names** — so those five would have appeared in both places, and a test asserting `unsupported()` is empty around a program that probes window state (`CSI 21 t` is ordinary) would newly fail.

The commit allows all eight, single-parameter as `seq.rs` requires, so `CSI 18;0t` — which the responder does not claim — is still recorded. I extended your test to pin both new edges and rewrapped the CHANGELOG entry, which was 153 columns against a file that wraps at about 74.

Title-stack behaviour is exactly as you wrote it: 22 and 23 still reported, and `queries.rs` still passes 23/23.

PR #428, opened 2026-09-16
test: cover signal and mouse wire tables

### LINKED ISSUE #400: Three hand-written wire tables have almost no coverage: six of seven signals, two mouse buttons and two wheel directions are never sent by any test
**Today** — three small tables translate a public enum into the bytes or
number that leave termlens, and a transposed arm in any of them would ship
silently.

```console
$ grep -rho 'Signal::[A-Z][a-z]*' crates/termlens/tests/ | sort -u
Signal::Term
$ grep -rc 'MouseButton::Middle' crates/termlens/tests/
0
$ grep -rc 'Scroll::Right' crates/termlens/tests/
0
```

`Signal::Term` is the only one of seven variants any test sends. All of
them behave correctly today, and each is already distinguishable through
the exit status, so the tests are writable as they stand:

```
Signal::Int  -> signal()=Some("Interrupt: 2")   Signal::Usr1 -> Some("User defined signal 1: 30")
Signal::Hup  -> Some("Hangup: 1")               Signal::Usr2 -> Some("User defined signal 2: 31")
Signal::Quit -> Some("Quit: 3")                 Signal::Kill -> Some("Killed: 9")
```

and the mouse codes encode correctly too, middle and right included, along
with all four wheel directions.

**Why it is worth fixing** — `signal()` is the graceful-shutdown API, and
the one thing that can be wrong about it is which signal actually left the
process. The mouse codes are the same shape of hand-written table, where a
middle-versus-right or a 66-versus-67 swap is invisible to the suite. Both
are cheap to pin and neither needs any new machinery.

**Fix** — two table-driven tests, and they can be two commits or two PRs:

1. In `crates/termlens/tests/process.rs`, a Unix-only test spawning the
   `emit` fixture, sending each `Signal` variant and asserting the exit
   status names it. Note that the macOS spellings carry a `: N` suffix, so
   assert on the stem rather than the whole string. Consider leaving
   `Signal::Quit` out if core files on CI are a concern, and say so.
2. In `crates/termlens/tests/input.rs`, extend the existing SGR wire test
   to a table over all three mouse buttons and all four wheel directions.

Mark the signal test `#[cfg_attr(windows, ignore = "…")]` as the other
signal tests are.

**Done when**
- [ ] Every `Signal`, `MouseButton` and `Scroll` variant is exercised and
      asserted by name.
- [ ] Swapping two arms of either table makes a test fail — check this by
      hand before trusting it.

---

Rayan-and-beyond:
## What & why

Closes #400.

Add table-driven coverage for every public `Signal`, `MouseButton`, and `Scroll` variant. Signal tests assert the host's signal description by stable name stem. Mouse and wheel cases use distinct coordinates so each enum-to-wire mapping is independently pinned.

The existing modifier and drag coverage stays in the same SGR wire test.

## Validation

- `cargo test -p termlens --test process every_signal_variant_reaches_the_child -- --exact`
- `cargo test -p termlens --test input buttons_modifiers_drag_and_horizontal_wheel_reach_the_wire -- --exact`
- Mutation proof: swapping `MouseButton::Middle` and `MouseButton::Right` codes makes the input test fail on the middle-button assertion
- Mutation proof: swapping `Signal::Int` and `Signal::Hup` mappings makes the process test fail with `Int: killed by signal: Hangup`
- `git diff --check`

Full workspace Clippy was not run locally because the shared development VM remains at its disk-headroom limit.

## Checklist

- [x] `CHANGELOG.md` updated under `[Unreleased]` (test-only change; no user-facing changelog entry needed)


vyncint (approved):
Thanks! Merging.

Both tables bite, which is the whole point of a test like this. I swapped `MouseButton::Middle` and `Right` in the source and the mouse test failed printing the entire wire, which is a genuinely good diagnostic. I swapped `Signal::Int` and `Hup` and the signal test failed with `Int: killed by signal: Hangup: 1`.

Giving every button and wheel direction its own coordinate is what makes each arm independently pinned — a shared coordinate would have let a middle-versus-right swap hide behind a `contains` check. And widening the terminal to 300 when the longer wire stopped fitting, rather than trimming the assertions to fit the old width, shows you read the failure instead of pattern-matching it.

On the gating: you used `#[cfg(unix)]` where the issue suggested `#[cfg_attr(windows, ignore)]`. The issue was wrong about the precedent — both neighbouring signal tests in that file use `#[cfg(unix)]`, and `Signal` itself is imported under it. You matched the file, which is the right call.

One thing the issue asked for that the PR does not address: it asked you to consider leaving `Signal::Quit` out over core dumps, and to say either way. I checked rather than leave it hanging — no core file appeared, `ulimit -c` is 0 on these runners, and Ubuntu pipes its core pattern to apport, so keeping `Quit` in is fine. Worth a sentence next time the issue asks you to make a call: saying "I considered it and here is why not" is as useful as the code.

`stress.yml` also mattered here, since this adds seven PTY spawns. The batch containing it went 15 legs green.

PR #429, opened 2026-09-16
fix(graphics): reject indirect kitty payloads

### LINKED ISSUE #402: A kitty file or shared-memory transmission decodes its own filename as pixels
**Today** — kitty's `t=` key says the payload body is a *path* or a
shared-memory name rather than pixel data. termlens neither records `t` nor
checks it, so the body is base64-decoded and handed back as an image:

```
t=t, f=32, s=1, v=1, body = base64("/tmp/x")
  decode() = Ok(1x1), pixel(0,0) = Some([47, 116, 109, 112])
```

`[47, 116, 109, 112]` is the ASCII of `/tmp`. A larger declared size gets a
diagnosis that blames the wrong thing instead:

```
t=f, f=32, s=64, v=64, body = base64("/etc/passwd")
  decode() = Err(Malformed("fewer bytes than the declared size needs"))
```

`GraphicsPayload` exposes no way to tell: there is no accessor for `t`, and
`Debug` does not mention it.

**Why it is worth fixing** — `kitty +kitten icat` uses temp-file and
shared-memory transmission by default, so this is the common path for a
real image rather than a synthetic one. A small declared size — an icon, a
cropped tile — produces a silently wrong `Ok`, which is exactly the failure
the module argues against in its own docs: an error naming the reason
rather than something a test could mistake for a real image. termlens
correctly never touches the disk; it simply does not say so.

**Fix** — in `crates/termlens/src/graphics.rs`, record `t` where the other
kitty keys are read, add a `GraphicsPayload` accessor for it (additive),
and return the existing `DecodeError::Unsupported` variant for `t=f`,
`t=t` and `t=s`. No new error variant is needed.

**Done when**
- [ ] `decode()` on a `t=f`, `t=t` or `t=s` payload is
      `Err(DecodeError::Unsupported(_))` whatever the declared size says.
- [ ] A unit test covers all three, plus `t=d`, which must still decode.
- [ ] `docs/LIMITATIONS.md` names file and shared-memory transmission in
      the graphics section, beside the `f=100` entry.

---

Rayan-and-beyond:
## What & why

Closes #402.

Record kitty's `t=` transmission medium on `GraphicsPayload` and expose it through `transmission()`. `decode()` now reports file, temporary-file, and shared-memory payloads as unsupported before interpreting their body as pixel bytes. Direct `t=d` data keeps decoding normally.

`Debug` includes a present `t=` value, and the graphics limitations document names the unsupported indirect transmission modes.

## Validation

- `cargo test -p termlens --lib --features decode graphics::tests::non_direct_kitty_transmissions_are_not_decoded_as_pixels -- --exact --nocapture`
- `cargo test -p termlens --lib --features decode graphics::tests -- --nocapture` — 24/24 passed
- `git diff --check`

Full workspace Clippy was not run locally because the shared development VM had about 20 MiB root-disk headroom after the focused Rust validation.

## Checklist


vyncint (approved):
Thanks! Merging.

The test is the strongest part: it drives the real builder through `kitty_payload` rather than hand-constructing a `GraphicsPayload`, so a broken `t=` parse fails here too, and it asserts the positive case — `t=d` decoding to its exact pixels — in the same function. That positive half is what catches the over-application later.

I reverted your guard and the test failed; with it in place, the issue's two literal reproductions both come back as `Unsupported` naming the medium, including the `t=f` case that previously blamed a short payload. You were also honest in the PR body about not running full clippy locally, which is the right instinct — it did fail, and I fixed it.

I pushed three commits:

**`9ec1abd`** does three things. Clippy first: `[b'f', b't', b's']` trips `byte_char_slices` under `-D warnings`, which every clippy job uses.

Then the accessor. `transmission()` returned `Option<u8>`, so a caller writes `== Some(b'f')`. Every other enumerated kitty key in this module is an enum with an `Other` arm and a `Display` — `GraphicsFormat` is the model — and 0.11 is the stability candidate, so whatever shape ships here is the shape we keep until 1.0. I added `GraphicsTransmission`, exported it, and gave it the same `t=f` display form your `Debug` already printed. Changing it later would need a new candidate, which is why I did it now rather than filing a follow-up.

Third: an unknown medium was still decoded as if direct. `f=` already refuses what it does not know, and guessing is precisely what turned a path into pixels, so `t=` now refuses too.

**`f7d35f1`** fixes a broken intra-doc link I introduced myself — `GraphicsPayload::decode` is behind the `decode` feature, so linking it fails `cargo doc` with default features. Plain backticks, as the `f=100` doc already uses for the same method.

Your refusal strings name the medium precisely and reuse `DecodeError::Unsupported` instead of inventing a variant, which is exactly what the issue asked for.

PR #430, opened 2026-09-16
fix(emu): clamp pending-wrap cursor position

### LINKED ISSUE #401: A program that fills the last column leaves the cursor at column `cols`, so the screen `inspect` prints is one `render` and `diff` refuse to read
**Today** — filling a row to its right margin leaves the backend in the
pending-wrap state, and termlens reports that as column `cols` — one past
the grid. The snapshot then fails termlens's own parser:

```console
$ termlens inspect --size 40x3 sh -c "printf '%040d' 0" > f2.snap
$ head -1 f2.snap
size: 40x3  cursor: 0,40
$ termlens render --text f2.snap
termlens: f2.snap: could not parse a saved screen: line 1: cursor 0,40 is outside the 40x3 grid
$ echo $?
2
$ termlens diff f2.snap f2.snap
termlens: f2.snap: could not parse a saved screen: line 1: cursor 0,40 is outside the 40x3 grid
```

One character short and it is fine:

```console
$ termlens inspect --size 40x3 sh -c "printf '%039d' 0" | head -1
size: 40x3  cursor: 0,39
```

The same column reaches two more places: a cursor-position report answers a
column that does not exist, and a graphics payload transmitted at the
margin reports a placement that is not a cell.

**Why it is worth fixing** — a full-width rule, status bar or progress bar
is ordinary TUI output, so this is not an exotic screen.
[#340](https://github.com/vyncint/termlens/issues/340) fixed the other
reason `inspect` output would not round-trip; this is a second,
independent one, on clean stdout. No real terminal reports a column past
its own width — xterm leaves the cursor at the last column and tracks
pending wrap as a separate flag.

**Fix** — clamp the reported column to the last cell in
`crates/termlens/src/emu/vt100.rs`, where the backend's cursor position is
read. One place fixes the round trip, the position report and the
placement together.

The conservative alternative is to accept `col == cols` in the parser and
document a pending-wrap column instead. That leaves the position report
answering a column past the width, so say in the PR why you chose it if you
go that way.

**Done when**
- [ ] `termlens inspect --size 40x3 sh -c "printf '%040d' 0"` piped into
      `termlens render --text` exits 0.
- [ ] A test in `crates/termlens/tests/queries.rs` pins the reply to a
      cursor-position report after exactly `cols` printable characters.
- [ ] A test round-trips a screen whose last column is written, from
      `to_string()` through `Screen::parse`.

---

Rayan-and-beyond:
## What & why

Closes #401.

Normalize vt100's pending-wrap cursor state to the last real grid cell when termlens exposes the cursor. That keeps saved screens parseable, cursor-position reports within the declared width, and graphics placement on an actual cell.

The regression fills the right margin exactly, round-trips the resulting saved screen through `Screen::parse`, and pins the CPR reply at column 10 for a 10-column terminal.

## Validation

- `cargo test -p termlens --lib emu::vt100::tests::a_pending_wrap_cursor_stays_inside_the_grid_and_round_trips -- --exact --nocapture`
- `cargo test -p termlens --test queries cursor_position_at_the_right_margin_reports_the_last_column -- --exact --nocapture`
- `cargo test -p termlens --test queries -- --nocapture` — 24/24 passed
- `git diff --check`

I did not build the CLI acceptance pipeline or run the full workspace Clippy matrix locally because the shared development VM remains disk-constrained; the focused round-trip and responder paths exercise the underlying fix directly.

## Checklist


vyncint (approved):
Thanks! Merging.

The diagnosis is right and the one-line clamp is the fix the issue preferred, wired through all three call sites — the snapshot, the cursor-position report and the graphics placement — via a single helper rather than three copies. Both new tests bite, and `parse(to_string()).to_string() == to_string()` is the right shape for a format invariant.

I also confirmed the fix composes with #375: on main a live 4x2 screen filled to the margin serialises to JSON its own reader then refuses. With your clamp the cursor comes back as `(0, 3, true)` and both the JSON and text round trips pass. That interaction is why this went in before #408.

I pushed `2db30d4` for two things the compiler cannot see.

`fn cursor_position` landed between `apply_tabs`'s doc comment and `apply_tabs`, so 22 lines about feeding staged bytes and `HT` semantics ended up documenting a four-line clamp helper, and `apply_tabs` was left bare. This is the second time in this batch a new item has stolen a neighbour's doc comment — easy to do, invisible to CI. I moved the helper to the end of the inherent impl and gave it a comment of its own.

The second is a real regression. `apply_tabs` read the clamped column, and it routes `HTS` and `TBC` through it. `TabStops::set` drops an out-of-range column deliberately — its own doc says clamping would set a stop the application did not ask for — so `HTS` at the margin, which used to be dropped, started creating a stop at the last cell, and that stop outlives the wrap once the grid widens. Measured: `HTS` at the margin of a 10-column grid, widened to 20, then `CHA 1` and a tab now lands on the default stop at column 8, where the clamped read put it at 9. `apply_tabs` reads the raw column again; the snapshot, the report and graphics placement still use the clamped helper.

Your CHANGELOG entry was 172 columns on one line; I rewrapped it and noted the tab-stop exception. The whole batch also passed `stress.yml`, 15 legs green.

PR #431, opened 2026-09-16
fix(recording): emit asciicast resize events

### LINKED ISSUE #398: `to_asciicast` writes the first frame's geometry into the header and never emits a resize event, so a recording that spans a resize replays at the wrong size
**Today** — the header takes its width and height from the first frame, and
the event loop emits only output events:

```
frame sizes: [(40, 10), (90, 20)]
header     : {"version": 2, "width": 40, "height": 10, "timestamp": …, "duration": …, "title": "…"}
event codes: ["o", "o"]
```

The second event repaints a 90x20 grid into a player that was told the
terminal is 40x10, so `asciinema play` and `agg` wrap and clip it. The
asciicast v2 format has an event for exactly this: `[time, "r", "COLSxROWS"]`.

**Why it is worth fixing** — `write_asciicast` is the documented way to
attach a session to a bug report, and a responsive-layout bug is precisely
the kind of thing you attach a recording of. It is the one recording that
renders wrong.

**Fix** — in `crates/termlens/src/terminal.rs`, track the previous frame's
size while iterating and emit an `r` event immediately before the output
event whenever it differs. This is purely additive to the exported
document; no public API changes.

**Done when**
- [ ] A recording spanning a resize produces event codes `["o", "r", "o"]`
      with the new geometry as the event's data.
- [ ] A fixed-size recording still produces only output events, and the
      existing tests in `crates/termlens/tests/record.rs` pass unchanged.
- [ ] A test parses the resize event back out of the document.

---

Rayan-and-beyond:
## What & why

Closes #398.

Track frame geometry while exporting asciicast v2 and emit an `r` event immediately before the first output frame at each new size. Fixed-size recordings keep their existing output-only event stream.

The regression records two synchronized frames across a real `Terminal::resize`, parses the exported events as JSON, and pins the `o, r, o` sequence plus `60x8` resize data.

## Validation

- `cargo test -p termlens --test record the_asciicast_emits_a_resize_before_the_first_frame_at_new_geometry -- --exact --nocapture`
- `cargo test -p termlens --test record -- --nocapture` — 8/8 passed
- `git diff --check`

Full workspace Clippy was not run locally because the shared development VM remains disk-constrained.

## Checklist


vyncint (approved):
Thanks! Merging — this is the last of the twenty-two.

Seeding `previous_size` from the first frame rather than from `None` is the detail that keeps the header and the event stream consistent. A `None` seed would have emitted a redundant resize event before every recording's first frame, duplicating what the header already says. Easy to get wrong, and you got it right without being asked.

Parsing the events back as real JSON instead of substring-matching the document is what makes the test survive a formatting change in the exporter. I dumped the whole document and checked it against the v2 spec: line 1 is a single object, every later line is an array, the resize event carries exactly `COLSxROWS`, and it is ordered immediately before the output at the same instant. Reverting your change fails the test on the event count.

I pushed `dc91cda` for the part CI cannot see. `to_asciicast`'s own rustdoc still promised "a header line, then one event per frame" — which this branch is precisely what makes false. It now says that a frame whose geometry differs from the one before it is preceded by a resize event, and why that matters for replay. Your CHANGELOG entry was 143 columns on one line; rewrapped, with the consequence spelled out.

The batch containing this passed `stress.yml`, 15 legs green.

PR #432, opened 2026-09-17
docs: refresh bug report form

### LINKED ISSUE #371: The bug-report form asks for a version with `0.5.0` as its example, and never asks which features were enabled
**Today** — the version field of `.github/ISSUE_TEMPLATE/bug_report.yml`
carries a placeholder six minor releases behind the published crate:

```console
$ grep -A4 'id: version' .github/ISSUE_TEMPLATE/bug_report.yml | grep placeholder
      placeholder: "0.5.0 (or git SHA)"
$ termlens --version
termlens 0.11.1
```

The form asks for the OS and for `TERM`, and never for the feature set —
though `decode`, `regex` and `serde` are optional, the crate is built four
different ways in CI for exactly that reason, and three past bugs
([#161](https://github.com/vyncint/termlens/issues/161),
[#226](https://github.com/vyncint/termlens/issues/226),
[#324](https://github.com/vyncint/termlens/issues/324)) existed only in
one configuration.

**Why it is worth fixing** — a placeholder is read as a suggestion of what
the maintainers expect to see, and this one suggests a version nobody
should still be running. The missing feature question costs a round trip
on any report that turns out to be configuration-specific, which is the
class hardest to reproduce from a screen dump.

**Fix** — in `bug_report.yml`, update the placeholder to the current
release, and add one optional field for the feature set with a placeholder
such as `default + decode` or the `Cargo.toml` line. While you are in the
file, consider suggesting `termlens inspect` in the "Expected vs actual"
description as the way to produce an attachable screen — the CLI exists
now and the form predates it.

**Done when**
- [ ] The version placeholder names a current release.
- [ ] The form asks which features were enabled, not required.
- [ ] The rendered form still submits with only the required fields.

---

Rayan-and-beyond:
## What & why

Closes #371.

Updates the bug-report version example to 0.11.1 and adds an optional enabled-features field so configuration-specific reports include the information needed to reproduce them.

## Checklist

- [x] Linked an issue
- [x] Validation added/run for the change: checked required fields remain required and the new features field is optional
- [x] `git diff --check`
- [x] All commits are signed off
- [x] No AI attribution trailers
- [ ] `CHANGELOG.md` updated — issue-template metadata only; no shipped crate behavior changed
- [x] No snapshot changes


vyncint (approved):
Thanks! Merging.

Both boxes met. The placeholder now names the current release, and the new feature field is optional rather than required — I parsed the YAML and confirmed the required set is unchanged at five, so the form still submits with only the fields it always demanded.

Two things done well. You put the feature question immediately after the version field rather than at the end: those are the two facts a triager needs together, and it reads as one question. And you matched the file's existing style exactly — same block indent, quoted placeholder, `description` as a sentence, and `validations.required` spelled out rather than omitted, the way the neighbouring optional field does it.

You left out the issue's `termlens inspect` suggestion, which was a "consider" rather than a Done-when box. That is a defensible call and the diff is now exactly the two boxes and nothing else. I will pick the other one up separately: the `expected` field still says "include the screen dump **if you have one**", which tells a reporter nothing about how to make one.

One note for me rather than you: nothing keeps that placeholder current, which is how it drifted six minor releases. That belongs in the release checklist, not in this pull request.

PR #433, opened 2026-09-17
fix(ci): include script gates in docs check

### LINKED ISSUE #383: `gates-listed` only extracts `cargo` steps, so the ten script gates are unchecked — and two of them are missing from CONTRIBUTING §1 right now

Rayan-and-beyond:
## What & why

Closes #383.

Extends `gates-listed` to cover single-line `.github/scripts/` and `tools/` CI gates alongside Cargo commands. Adds the two currently missing script gates to the contributor checklist and reports cargo/script counts separately.

## Checklist

- [x] Linked an issue
- [x] Tests added/updated for the change: mutation check proves removing `check-skill-snippets.sh` from CONTRIBUTING makes the gate fail
- [x] `bash .github/scripts/check-ci-gates-listed.sh`, `bash -n`, `shellcheck`, and `git diff --check` are clean
- [x] All commits are signed off
- [x] No AI attribution trailers
- [x] No user-facing changelog entry needed for a CI/docs gate fix
- [x] No snapshot changes


vyncint (changes requested):
This is a good fix and the bite test holds — I removed five different script gates from CONTRIBUTING §1 in turn and the gate went red on every one, naming the missing line. Splitting the success message into `19 cargo commands, 12 script gates` is the right touch: a reader can now *see* that the script gates are covered, which is exactly why the old bug went unnoticed.

**One thing to fix before merge.** The two new counters on lines 16-17 of `.github/scripts/check-ci-gates-listed.sh` run before the comparison loop, and `grep -c` exits 1 when it matches nothing. Under `set -euo pipefail` that means the script now dies silently on any workflow that has no script gates — and that is reachable through the script's own documented `[ci.yml]` argument:

```console
$ .github/scripts/check-ci-gates-listed.sh .github/workflows/release.yml CONTRIBUTING.md
$ echo $?
1
```

No output at all. Before your branch the same call reported the real problem:

```console
::error::CI runs `cargo publish -p termlens --locked` but CONTRIBUTING.md §1 does not list it
```

The same happens for `commit-policy.yml`, `install.yml`, `windows.yml` and `stress.yml`. CI cannot catch it because CI only ever passes the default `ci.yml`, which has both kinds of gate.

The fix is one token on each line:

```bash
cargo_count="$(grep -c '^cargo ' <<<"$ci_cmds" || true)"
script_count="$(grep -Ec '^(\.github/scripts/|tools/)' <<<"$ci_cmds" || true)"
```

Nothing else needs to change.

**Optional, take it or leave it:** the new `.github/scripts/check-ci-gates-listed.sh` line you added to §1 sits inside the feature-isolation group, while the gate itself runs in the `gates-listed` job beside the selftests. Moving it down next to `tools/link-gate-selftest/run.sh` would keep §1's job-by-job grouping intact.

Ping me when that is in and I will merge. Worth knowing: this and #441 merge cleanly in both directions, and once both land the success line ticks to `19 cargo commands, 13 script gates` — your widening immediately covers the gate #441 adds.

Rayan-and-beyond:
Fixed the zero-match `grep -c` case and pushed it. The default gate still passes, and the `release.yml` example now exits 1 with the expected missing-gate error instead of failing silently.

vyncint (approved):
Verified locally beyond CI: the gate fails for all three shapes it now claims — a dropped `cargo` gate (`cargo fmt --all --check`), a dropped `.github/scripts/` gate (`check-skill-snippets.sh`) and a dropped `tools/` gate (`link-gate-selftest/run.sh`) each go red naming the missing line, and the branch is green with 19 cargo commands and 12 script gates.

Worth recording for the next person: the gate is deliberately one-directional (CI → doc). A doc line with no `- run:` counterpart in ci.yml — `cargo deny check`, which runs via `uses: cargo-deny-action` — is not watched by it, before this change or after. That is not a gap this PR introduces and the split counts now make it visible.

PR #434, opened 2026-09-17
fix(screen): serialise the style booleans in SGR order

### LINKED ISSUE #381: DESIGN §3 says the JSON style booleans are in SGR order; `reverse` is written before `blink`
**Today** — the specification says it twice:

```console
$ grep -n 'SGR order' docs/DESIGN.md
544:   `underline`, `blink`, `reverse`, `conceal`, `strikethrough` — SGR order,
773:  booleans in SGR order; a colour is `"default"`, `{"indexed": 4}` or
```

The emitted document says otherwise, because `serde`'s derive follows
declaration order and `Style` declares `reverse` first:

```json
"bold":false,"dim":false,"italic":false,"underline":false,
"reverse":false,"blink":true,"conceal":false,"strikethrough":false
```

SGR order is bold 1, dim 2, italic 3, underline 4, **blink 5, reverse 7**,
conceal 8, strikethrough 9 — which is what the text format's `styles:`
block writes and what the ANSI rendering emits. The JSON is the only one of
the three that disagrees with the spec.

**Why it is worth fixing** — DESIGN §3 is the frozen specification of JSON
format 1, and `docs/STABILITY.md` names that shape as promised. Someone
writing a reader in another language against the spec gets a different
field order than termlens produces, and a reader of the spec cannot tell
which of the two is the bug.

**Fix** — one file, and either direction is defensible; say which you chose
and why:

1. Swap the `blink` and `reverse` declarations in `Style`
   (`crates/termlens/src/screen.rs`). `serde` is key-based, so no existing
   document stops deserialising; only the emitted key order and the `Debug`
   order move, and both are unpromised.
2. Correct the two sentences in `docs/DESIGN.md` to describe the order
   actually emitted.

**Done when**
- [ ] The spec and the emitted document agree.
- [ ] A test pins the order of the eight boolean keys in the serialised
      document, so the two cannot drift apart again.
- [ ] `crates/termlens/tests/compat.rs` still reads every corpus JSON.

---

maskjelly:
Closes #381.

**Which direction, and why:** the swap (direction 1). DESIGN §3 is the frozen specification of JSON format 1 and `docs/STABILITY.md` names that shape as promised; the text format's `styles:` block and the ANSI rendering already follow SGR order, so the JSON was the lone dissenter. Correcting a frozen spec to bless an accidental declaration order is the other way round.

**Changes:** `Style` declares `blink` before `reverse` (`crates/termlens/src/screen.rs`), with a comment recording why declaration order is load-bearing here. Serde reads by key, so every existing document still deserialises — only the emitted key order and the `Debug` order move, both unpromised.

**The test pins the thing that drifted:** a `Value` forgets key order, so `the_style_booleans_serialise_in_sgr_order` reads it out of the emitted string and requires the eight positions to strictly increase in SGR order. With the declaration order reverted it fails (`reverse is out of SGR order in "style":{…"underline":true,"reverse":true,"blink":true…}`); with the fix it passes.

**Snapshot:** `export__json__styled_json.snap` changed — 120 cell styles, and each changed line is exactly the `reverse`/`blink` swap, verified by filtering the diff for anything that is not one of those two keys. It was accepted after reading that diff, not blind-accepted.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (48 suites, 0 failed), `cargo test -p termlens --all-features --test compat` (every corpus JSON still reads), plus the no-default-features clippy run since the new test is feature-gated.


vyncint (approved):
Thanks! Merging.

You picked the right one of the two options the issue offered. DESIGN §3 already specified SGR order, in two places, so the specification needed no edit — the code was the deviation. Fixing the code rather than the spec is the harder call and the correct one.

The new test is what makes this stick. It reads the **emitted string** rather than a `serde_json::Value`, and your rustdoc says why: a `Value` no longer remembers the order. That is the difference between a test that pins this and one that structurally cannot. The existing corpus test compares values and is therefore order-blind by construction, which is exactly why the drift survived until now — you found the gap and closed it at the right layer.

I checked the snapshot by hand, since a key-order diff is precisely the kind a blind `cargo insta accept` would wave through. Sorted line-for-line the two versions are identical, and parsing both and deep-comparing with key order ignored gives equality — so the change is order only, with nothing added, removed or re-valued. All eighteen corpus JSON files still carry the old order and all still read, which is the compatibility case rather than a no-op.

Reverting the two field declarations fails both the new test and the round-trip test, with a message that names the offending key.

I also want to note the comment you added above the field block. It explains why the order is load-bearing, and you placed it as a plain `//` above the `///` doc comment so it does not steal `bold`'s rustdoc. Two pull requests in this batch fell into exactly that trap; you avoided it deliberately.

One suggestion for the future rather than a change I am asking for: the entry says only the emitted key order moved, which is true, but the first thing an upgrading user meets is churn in their own committed JSON snapshots. A closing sentence saying so — one review pass, no value changes — would save them working it out. This pull request had to re-review 240 lines of snapshot for that very reason.

I left `skills/termlens/SKILL.md` alone. Its field list, like the README table, reads as "these exist" rather than as a declaration order, and the README deliberately groups the three later additions together.

PR #435, opened 2026-09-17
fix(cli): render accepts exactly one operand

### LINKED ISSUE #364: `termlens render` silently ignores every file operand but the last, so a typo renders the wrong screen
**Today** — `render`'s argument loop assigns `file = Some(arg)` for each
non-flag operand, so the last one wins and the others vanish. Measured with
the published 0.11.1 binary, on two saved screens whose only difference is
that one says `hi` and the other says `bye`:

```console
$ termlens render --text a.snap b.snap | sed -n 2p
bye
$ echo $?
0
```

The worst shape is with `--out`, where nothing is printed to look at:

```console
$ termlens render --svg --out shot.svg a.snap b.snap
$ echo $?
0
$ grep -o '>hi<\|>bye<' shot.svg
>bye<
```

`diff` does not behave this way: it collects into a `Vec` and destructures
`[a, b]`, so a third operand prints the usage and exits 2.

**Why it is worth fixing** — the operand is the whole input. A user who
writes two paths, or who leaves a stale path on the line after editing a
command, gets exit 0 and a rendering of a screen they did not name. In a CI
step that writes an attachment with `--out`, nothing ever reveals it. This
is the quiet-wrong-answer shape [#260](https://github.com/vyncint/termlens/issues/260)
was about, and `diff` already has the cure in the same file.

**Fix** — in `crates/termlens-cli/src/main.rs`, collect the operands of
`render` into a `Vec` as `diff` does and accept exactly one, printing
`RENDER_USAGE` and exiting 2 otherwise. Consider whether two format flags
(`--svg --text`) should also be refused rather than silently letting the
last win; if you leave that as is, say so in the PR so it is a decision
rather than an oversight.

**Done when**
- [ ] `termlens render --text a.snap b.snap` exits 2 and prints the usage,
      and `--out` creates no file in that case.
- [ ] One operand still behaves exactly as today, `-` included.
- [ ] `crates/termlens-cli/tests/cli.rs` covers the two-operand case, and
      `.github/scripts/check-cli-contract.sh` gains the same check beside
      its existing `diff - -` one.

---

maskjelly:
Closes #364.

`render`'s argument loop assigned `file = Some(arg)` per operand, so the last one won and the others vanished — `render --text a.snap b.snap` rendered b with exit 0, and with `--out` the wrong screen was written where nothing printed to reveal it.

**Fix:** operands are collected into a `Vec` and one refutable destructure accepts exactly one, printing `RENDER_USAGE` and exiting 2 otherwise — the shape `diff` in the same file already uses. The check runs before any file is read or stdin consumed, so `--out` creates no file and `-` cannot be half-consumed.

**Measured before/after** (two screens differing only in `hi`/`bye`):

```
before: render --text a.snap b.snap          → exit 0, prints "bye"
        render --svg --out shot.svg a.snap b.snap → exit 0, shot.svg says "bye"
after:  render --text a.snap b.snap          → exit 2, full usage on stderr
        render --svg --out shot.svg a.snap b.snap → exit 2, no shot.svg
        render --text a.snap                 → exactly as before, "-" included
```

**The format-flag question, answered explicitly:** `--svg --text` still lets the last flag win. The issue asks for the operand to be fixed and for the format conflict to be considered; refusing it would extend the documented CLI contract with a diagnostic the issue does not specify. It is the same silent-wrong-answer class as the operand and a reasonable follow-up if you want it.

**Tests:** a `cli.rs` case pins both operands-with-usage and `--out`-creates-nothing, plus that the count check precedes the read (`render --text - a.snap` never consumes stdin); with last-wins temporarily restored it fails (`left: Some(0), right: Some(2)`). `check-cli-contract.sh` gains the two checks beside its `diff - -` pair, so the published binary is held to them too.

**Stress note:** the new test spawns the binary, so CONTRIBUTING §3's stress trigger applies; I cannot dispatch workflows from a fork — could you run `stress.yml` on this branch?

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (cli.rs 17/17), `check-cli-contract.sh` PASS with the new lines.


vyncint (approved):
Thanks! Merging.

The test case I did not expect is the best one: `render --text - a.snap` asserting that stderr does **not** mention `<stdin>`. That pins the operand count being checked *before* the pipe is read, which is the part that actually makes `--out` safe. Nothing in the issue asked for it.

I drove the release binary through the whole space: two operands exit 2 with the usage; `--svg --out shot.svg a.snap b.snap` exits 2 and creates no file; three operands exit 2; zero operands exit 2. One operand is untouched, including `-` on both the stdout and the `--out` path, and `- a.snap` is refused without reading stdin.

Both halves bite. Restoring last-wins fails the new test, and — the part I care about more — it also fails the contract script, with `FAIL render, two operands exit 0, expected 2 with the usage` and the `--out` twin. A gate that cannot go red is a recurring defect in this repository, and yours can.

Answering the issue's optional `--svg --text` question explicitly, as a decision with a reason rather than silence, is exactly what was asked for.

One small correction to the body: you asked me to dispatch `stress.yml` because the new test spawns the binary. It does not need it — that test uses `std::process::Command` with piped stdin and adds no PTY spawns, so CONTRIBUTING §3's trigger does not apply. The rule is about tests that open pseudo-terminals, which is what makes the suite timing-sensitive. Saves a three-platform run.

PR #436, opened 2026-09-17
fix(cli): keep the partial-comparison note in the coloured diff

### LINKED ISSUE #365: The coloured `termlens diff` drops the line saying the comparison was partial, so a size change reads as a full diff
**Today** — when two saved screens differ in size, `ScreenDiff`'s plain
rendering says so on its second line. The coloured rendering the CLI paints
for a terminal keeps the first line and the rows, and throws that one away.
Measured with the published 0.11.1 binary on a 4x3 and a 6x2 screen:

```console
$ termlens diff --color never d1.snap d2.snap
size: 4x3 → 6x2   cursor: 0,0
compared over the 4x2 overlap; the rest is clipped
  0 │ab  │abXX
    │  ^^│  ^^
  1 │cd  │ZZ
    │^^  │^^

$ termlens diff --color always d1.snap d2.snap | sed 's/\x1b\[[0-9;]*m//g'
size: 4x3 → 6x2   cursor: 0,0
  0 │ab  │abXX
  1 │cd  │ZZ
```

The cause is in `colored()` in `crates/termlens-cli/src/main.rs`: it takes
`lines.next()` as *the* header — exactly one line — then rebuilds the row
block itself, and filters the tail with
`l.starts_with('…') || l.starts_with("styles:")`. The overlap note starts
with neither, and it is not a row, so nothing carries it.

**Why it is worth fixing** — that sentence is the one that says the answer
is incomplete: columns 4 and 5 of the right-hand screen and row 2 of the
left-hand one were never compared, and a cell that changed there is not in
this listing. It is dropped precisely where a human is reading — `--color
auto` turns colour on for a terminal and off for a pipe — so the CI log
tells the truth and the interactive session does not. The `^^` marker lines
are dropped on purpose (colour replaces them, and the code says so); this
line is not that.

**Fix** — in `colored()`, carry the header through until the first row
instead of assuming it is one line: take lines from the plain rendering
until one matches the row shape, or explicitly keep a line beginning
`compared over the`. Whichever you choose, make it survive a future
addition to the header rather than listing another prefix, and say in a
comment which lines the function owns and which it passes through.

**Done when**
- [ ] `termlens diff --color always` over two screens of different sizes
      prints the overlap note.
- [ ] The colouring of rows, the `…` line and the `styles:` lines are
      unchanged, and `no difference` still prints alone.
- [ ] `crates/termlens-cli/tests/cli.rs` pins it with a `--color always`
      case over two differently sized screens.

---

maskjelly:
Closes #365.

When two saved screens differ in size, `ScreenDiff`'s plain rendering says so on its second line — `compared over the 4x2 overlap; the rest is clipped`. The coloured rendering dropped it: `colored()` took `lines.next()` as *the* header, exactly one line, then filtered the tail for `…`/`styles:`, and the note matches neither.

That sentence is the one saying the answer is incomplete — a change in the clipped rows or columns is not in the listing — and it was dropped precisely where a human reads (`--color auto` is on for a terminal, off for a pipe), so the CI log told the truth and the interactive session did not.

**Fix:** the boundary between header and rows is detected by shape, not by counting lines — a row is `{row:>3} │…│…`, so everything before the first line with that shape passes through untouched and only the rows (which the function owns, rebuilt and painted cell by cell) are handled specially. A future header line is carried without adding another prefix to a list; the function's comment now says which lines it owns and which it passes through. Row colouring, the `…` line, `styles:` lines and `no difference` are unchanged.

**Measured after** (issue's exact case, ANSI stripped):

```
size: 4x3 → 6x2   cursor: 1,2
compared over the 4x2 overlap; the rest is clipped
  0 │ab  │abXX
  1 │cd  │ZZ
```

**Test:** a `cli.rs` case runs `diff --color always` over two differently sized screens (new fixtures under `tests/data/`) and asserts the exact note, the cell colouring (parsed from the PTY screen, same mechanism as the existing colour test), and that `no difference` still prints alone with no styled cell. Before the fix it is red on the missing note.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (48 suites, 0 failed), `check-cli-contract.sh` PASS, `check-candidate-statement.sh` still ok with the CHANGELOG touched.


maskjelly:
Windows caught something worth recording: the test asserted that the two changed blanks on the before side are painted red, and ConPTY does not carry a foreground-only run written around isolated spaces across the line boundary. The `windows` leg rendered the two blanks uncoloured and the *next* row's red run two cells wider (`3: 5-8 fg=1` instead of `3: 5-6 fg=1`), so the assertion is now pinned on non-Windows and the comment says why with the run number as evidence. The overlap note itself — the thing this PR fixes — renders on every platform and is still asserted everywhere. Force-pushed.

vyncint (approved):
Thanks! Merging.

Solving it by shape — the new `is_row_line` helper — rather than by adding another prefix to the tail filter is exactly what the issue asked for, and it means a future header line needs no change here at all. That is the difference between fixing this bug and fixing this class of bug.

I compared the old and new binaries byte for byte across seven invocations. Five are identical, including both same-size cases, both `no difference` cases and the plain rendering. Only the two coloured differently-sized runs differ, and the entire delta is the one line:

```
compared over the 4x2 overlap; the rest is clipped
```

That is the whole behavioural change, which is what I want to see in a fix to a painting routine. The row colouring, the unchanged-row line and the `styles:` lines all survive, and reverting `main.rs` fails the new test with the note absent from the screen it prints.

I also probed an edge the issue did not mention: two screens differing only in size, with no changed cell at all. There is no row line, so the header loop drains everything — and the output comes out header, note, unchanged-row count, in order. Strictly better than before, where the note was dropped.

Your two new data files are genuinely different sizes and match the issue's reproduction, and the Windows carve-out is documented with the run id and the ConPTY reason, while the note assertion itself still runs on every platform.

On `stress.yml`: the tests add two PTY spawns, so CONTRIBUTING §3 applies and I dispatched it. Worth telling you what happened, because it is instructive. The first run failed two Windows legs — `two_dozen_terminals_open_at_once` and `unanswerable_queries_turn_timeouts_into_diagnoses`. Neither is a file this branch touches, and the re-run of the identical tree went 15 for 15. Both are pre-existing flakes on a loaded Windows runner, not yours. I am filing them separately.

One cosmetic note: the "Measured after" block in your description shows `cursor: 1,2`, while the fixtures and the real output say `cursor: 0,0`. Description only.

Issue #372, opened 2026-09-16
The `inspect` example and `termlens inspect` are two implementations of one tool, and they now disagree on the exit code and on which stream the trailer uses


vyncint:
**Today** — `crates/termlens/examples/inspect.rs` is 202 lines that do what
the published `termlens inspect` does. Since the CLI crate arrived they
have drifted apart, on exactly the two points `docs/STABILITY.md` lists as
promised CLI behaviour. Measured on 0.11.1:

```console
$ target/release/examples/inspect /no/such/prog; echo "exit=$?"
inspect: failed to spawn `/no/such/prog`: Unable to spawn … (ENOENT: No such file or directory)
exit=1
$ termlens inspect /no/such/prog; echo "exit=$?"
termlens: failed to spawn `/no/such/prog`: Unable to spawn … (ENOENT: No such file or directory)
exit=2
```

```console
$ target/release/examples/inspect /bin/echo hi 2>/dev/null | tail -1
--- exited: exit code 0 ---
$ termlens inspect /bin/echo hi 2>/dev/null | tail -1

```

The example's own module docs still teach the old contract — "Exit code 1
means inspect itself could not run" — and the trailer on stdout is the
shape [#340](https://github.com/vyncint/termlens/issues/340) removed from
the CLI because it made `inspect … > file` produce a file the tool would
not read back. The example also never gained `--cwd` or `--version`.

**Why it is worth fixing** — two tools with one name and two contracts.
Someone reading `examples/` to learn the crate learns exit codes that
contradict the ones the CLI promises and `check-cli-contract.sh` enforces,
and `docs/HANDOFF.md:256` still sends people to
`cargo run --example inspect` rather than to the installed command.

**Fix** — pick one and say which in the PR:

1. **Align it.** Make the example exit 2 on its own failure, put the
   trailer on stderr, and correct its module docs. Smallest diff; keeps a
   worked example of driving `Terminal` from `main`.
2. **Retire it.** Delete the example and `crates/termlens/tests/inspect.rs`
   with it, point `docs/HANDOFF.md` at `cargo install termlens-cli`, and
   note it in the CHANGELOG. This also removes the `cargo build --example`
   that runs *inside* a test, which
   [#278](https://github.com/vyncint/termlens/issues/278) was a race in.

Either is a fine first PR; (1) is the safer one to start with.

**Done when**
- [ ] One `inspect` contract exists in the repository, and nothing
      documents the other.
[...]

PR #437, opened 2026-09-17
feat(cli): accept `--flag=value` on inspect

### LINKED ISSUE #366: `termlens inspect --size=120x40` is an unknown option, though `diff --color=` and `render --out=` both work
**Today** — `diff` and `render` each accept the `--flag=value` form beside
the two-argument one. `inspect` accepts none of them. Measured with the
published 0.11.1 binary:

```console
$ termlens diff --color=never a.snap a.snap
no difference
$ termlens render --text --out=/tmp/o.txt a.snap
$ echo $?
0

$ termlens inspect --size=100x30 /bin/echo hi
termlens: unknown option "--size=100x30" (try `termlens inspect --help`)
$ termlens inspect --timeout=1 /bin/echo hi
termlens: unknown option "--timeout=1" (try `termlens inspect --help`)
```

The same for `--idle=`, `--cwd=` and `--env=`. In
`crates/termlens-cli/src/main.rs`, `diff` and `render` each carry an arm
matching `other.starts_with("--color=")` / `"--out="`; `inspect`'s loop
matches whole flags only, so an `=` lands in the catch-all.

**Why it is worth fixing** — one tool, two conventions, and the one that
refuses is the command people type most. `--env=KEY=VALUE` is the form
someone copying a `docker run` line will reach for first, and the
diagnostic calls the flag unknown when it is the *spelling* that is
unsupported, which sends the reader to `--help` for a flag that is already
listed there.

**Fix** — in `inspect`'s flag loop, split a leading `--name=` before
matching, and feed the value to the same `take` helper the two-argument
form uses, so every flag gains the form at once and each keeps its own
diagnostic. `--env=A=b` must set `A` to `b`: split on the *first* `=` only.

**Done when**
- [ ] `--size=`, `--timeout=`, `--idle=`, `--cwd=` and `--env=` all work
      and mean exactly what their two-argument forms mean.
- [ ] A bad value still names the flag the user typed, e.g.
      `bad --size "12", expected e.g. 120x40`.
- [ ] `crates/termlens-cli/tests/cli.rs` covers at least `--size=` and
      `--env=KEY=VALUE`.

---

maskjelly:
Closes #366.

`diff --color=never` and `render --out=PATH` work; `inspect --size=120x40` was refused as an unknown option. One tool, two conventions — and the one that refused is the command people type most, with `--env=KEY=VALUE` being the spelling someone copying a `docker run` line reaches for first. The diagnostic also called the *flag* unknown when it was the spelling that was unsupported.

**Fix:** one split, not five special cases. A leading `--name=` is split before inspect's flag match and the value goes in front of the argument stream the existing helper reads, so every flag gains the form at once and each keeps its own diagnostic unchanged. The whole token stays in the unknown-option message, quoting what the user actually typed. `--env=A=b=c` splits on the first `=` only and sets `A` to `b=c`.

**Measured before/after:**

```
before: inspect --size=100x30 /bin/echo hi  → unknown option "--size=100x30", exit 2
after:  inspect --size=100x30 /bin/echo hi  → size: 100x30 … render, exit 0
        inspect --timeout=1 …                → exit 0
        inspect --env=A=b=c sh -c 'echo "[$A]"' → [b=c], exit 0
        inspect --size=12 …                  → bad --size "12", expected e.g. 120x40, exit 2
        inspect --cwd=/no/such/dir true      → bad --cwd "/no/such/dir", not an existing directory, exit 2
        inspect --nope=1                      → unknown option "--nope=1" (whole token kept), exit 2
```

**One deliberate widening, disclosed:** boolean flags now accept an ignored inline value (`--ansi=1` exits 0 where it previously said unknown option). That falls out of the one-mechanism split; say the word if you would rather booleans refuse a value and I will special-case them.

**Tests:** `--size=`/`--timeout=`/`--idle=` land, `--env=` sets `A` to `b=c` (first-`=` split), and a bad `--size=12` still names the flag with the existing wording. All three fail with the mechanism reverted.

**Stress note:** the new tests add PTY spawns; per CONTRIBUTING §3 I cannot dispatch `stress.yml` from a fork — happy to have you run it on this branch.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features`, `check-cli-contract.sh` PASS.


vyncint (changes requested):
The parity here is exact and I checked it character for character: `--size=`, `--timeout=`, `--idle=`, `--cwd=` and `--env=` all behave identically to their two-argument forms, `--env=A=b=c` splits on the first `=` as it should, and every diagnostic names the flag the user typed. Splitting before the match and feeding the inline value through the same `take` helper is the right shape — every flag gained the spelling at once and each kept its own message.

The `--cwd=` test building and canonicalising its own scratch directory, with a comment about the macOS `/tmp` symlink, is the kind of care that prevents a green-locally red-on-CI report.

**One thing to tighten before merge.** The split guard is `name.starts_with("--") && name != "--"`, so it splits *every* long flag, including the ones that take no value. The value is then dropped and the flag applied anyway:

```console
$ MYVAR=zzz termlens inspect --size 40x3 --inherit-env=nonsense sh -c 'echo "[$MYVAR]"'
size: 40x3  cursor: 1,0
[zzz]
$ echo $?
0
```

On main that is `exit 2, unknown option "--inherit-env=nonsense"`. The same now happens for `--ansi=nonsense`, `--version=nonsense` and `--help=nonsense`, all of which become exit 0. And `render`, the arm this mirrors, refuses the shape:

```console
$ termlens render --svg=1 a.snap
termlens: unknown option "--svg=1" (try `termlens render --help`)
```

So it makes `inspect` looser than the convention it is adopting, and it does it on `--inherit-env`, which decides which screen you get — a silent wrong answer of the class #364 and #365 were about.

The smallest fix is to split only the flags that take a value and let the rest fall through to the existing catch-all, which then reports the whole token:

```rust
let (name, inline) = match flag.split_once('=') {
    Some((name, value))
        if matches!(name, "--size" | "--timeout" | "--idle" | "--cwd" | "--env") =>
    {
        (name, Some(value))
    }
    _ => (flag.as_str(), None),
};
```

A one-line case in `cli.rs` asserting `--inherit-env=nonsense` exits 2 would keep it that way.

**Two other things worth knowing.**

#443 is open and closes #372, which was about the `inspect` example and `termlens inspect` having drifted into two contracts. Today they agree that `--size=60x3` is an unknown option. The moment this merges, the CLI accepts it and the example still refuses — a fresh divergence on the same axis #443 just closed. Whichever of the two lands second should carry this parser into `crates/termlens/examples/inspect.rs`. Since #443 is ready now, the cheaper order is for it to go first and for this PR to mirror the split into the example.

The tests add three PTY spawns, so CONTRIBUTING §3 asks for a `stress.yml` run. A fork cannot dispatch it — say so in the body and I will run it.

maskjelly:
Both points addressed and force-pushed.

**The split is now only for the flags that take a value** — the five the issue names, using your snippet — so a value on a flag that takes none falls through to the catch-all and reports the whole token. Verified the exact case from your review:

```console
$ termlens inspect --size 40x3 --inherit-env=nonsense sh -c 'echo "[$MYVAR]"'
termlens: unknown option "--inherit-env=nonsense" (try `termlens inspect --help`)
$ echo $?
2
```

A `cli.rs` case pins it (`inspect_refuses_a_value_on_a_flag_that_takes_none`), so `--ansi=nonsense`, `--version=nonsense` and `--help=nonsense` are exit 2 again as on main.

**The example carries the same parser.** #443 landed first, so this PR mirrors the split into `crates/termlens/examples/inspect.rs` — same five flags, same refusal for the rest `(unknown option "--inherit-env=nonsense"` with the example's own `--help` suffix) — and `tests/inspect.rs` now pins both the `--size=12x3` spelling working and the value-less refusal, so the pair agree on the axis #443 closed.

Gates re-run on the rebased tree: fmt, clippy `-D warnings`, full workspace suite (48 result blocks, none failed), CLI contract script PASS.

**Stress:** the three new PTY spawns still stand, and a fork cannot dispatch `stress.yml` — could you run it on this branch before merging?

vyncint (dismissed):
Approved. One split in front of the flag match, rather than five special cases, is the right shape — every flag gains the spelling at once and each keeps its own diagnostic, which the `bad --size "12"` test pins.

Checked the parse by hand as well as by CI: `--size=60x3`/`--timeout=`/`--idle=` land, `--env=A=b=c` sets `A` to `b=c` (first `=` only), `--cwd=` reaches `current_dir`, and `--inherit-env=nonsense` is refused with the whole token named. Mirroring it into `examples/inspect.rs` matters — #443 had just realigned that example with the CLI, and this keeps them from drifting apart again.

**Your PR description is stale in one place, in your favour.** It says booleans now accept an ignored inline value (`--ansi=1` exits 0). The code does not do that: the split is restricted to the five value-taking flags, so `--ansi=1` falls to the catch-all and is refused — and you have a test asserting exactly that. The final behaviour is the better one; only the prose describes the earlier version.

**Stress: 15/15.** You could not dispatch it from a fork, so I ran it on your branch merged with #439 onto current main — the tree that will exist after both land — across all three OSes and five thread counts. That is the CONTRIBUTING §3 requirement for new PTY spawns and it is met.

Rebased onto main to resolve a CHANGELOG conflict: #447 added an `### Added` section an hour before this, and both bullets now sit under the one heading.

PR #438, opened 2026-09-17
ci: audit the whole .github tree, not just the workflows

### LINKED ISSUE #389: The zizmor job audits `.github/workflows/` only, so the composite action this repository ships to other repositories is never audited
**Today** — the workflow audit is scoped to one directory:

```console
$ grep -n 'zizmor' .github/workflows/ci.yml | grep persona
  pipx run zizmor==1.29.0 --persona=pedantic .github/workflows/
$ grep -n 'zizmor' CONTRIBUTING.md | head -1
57:pipx run zizmor==1.29.0 --persona=pedantic .github/workflows/
```

`.github/actions/report/action.yml` sits outside that path. It is the one
file other repositories execute with `uses:`, it interpolates its inputs
into `bash` steps, and it pins `actions/upload-artifact` by SHA. zizmor
parses it happily when asked:

```console
$ zizmor --persona=pedantic --no-online-audits .github/actions/report/action.yml
completed .github/actions/report/action.yml
No findings to report. Good job!
```

**Why it is worth fixing** — it is clean today, so this is preventive
rather than a live defect. But `SECURITY.md` describes the pinning and
input-handling rules in terms that read as covering the action, and the
action is the only artifact here that runs inside someone else's
repository. An unaudited change to it is the one that would matter most.

**Fix** — point the audit at `.github/` instead of `.github/workflows/` in
`ci.yml` and in CONTRIBUTING §1, and check that the accepted findings in
`.github/zizmor.yml` still cover what the wider scope turns up.

**Done when**
- [ ] The job audits `.github/actions/report/action.yml`.
- [ ] Introducing a `${{ inputs.… }}` interpolation inside a `run:` block of
      that action makes the job go red — verify by hand, then revert.
- [ ] CONTRIBUTING §1 names the same path as the workflow.

---

maskjelly:
Closes #389.

The zizmor job audited `.github/workflows/` only, so `.github/actions/report/action.yml` — the one artefact this repository ships for other repositories to execute, with inputs interpolated into `bash` steps and an upload action pinned by SHA — was never audited. It is clean today; this is preventive, and it is the file where an unaudited change would matter most.

**Scope change:** the audit points at `.github/` now, in `ci.yml` and in CONTRIBUTING §1 (the two must name the same path; `check-ci-gates-listed.sh` still passes, and the zizmor pin it greps is unchanged).

**What the wider scope turned up, and what I did with it:**

```
old scope .github/workflows/ → No findings to report (16 ignored, 1 suppressed)
new scope .github/          → 19 findings (16 ignored, 1 suppressed, 2 safe fixes)
```

The two new ones are real `dependabot-cooldown` findings in `.github/dependabot.yml` (no cooldown; GitHub's default is three days, zizmor wants seven). I accepted them in `.github/zizmor.yml` with a comment rather than fixing them here, because changing the dependency-update policy is a separate decision from pointing the audit at `.github/` — the one-line fix and the entry are both in the diff discussion if you would rather take the former.

**Hand-red verified, as the Done-when asks:** adding `echo "red-check: ${{ inputs.name }}"` inside the action's `run:` block turns the new scope red with `error[template-injection] … may expand into attacker-controllable code` (exit 14) while the same file under the old scope stays green — then reverted byte-identically. No findings live in the action today; the point is that the job would now catch one.

**Not done:** the job's display name still says "Audit workflows at pedantic level" — it names no path, and renaming it touches the required-check configuration, so it is left for a separate change if wanted.

**Gates:** `check-ci-gates-listed.sh` pass, YAML parses for all three touched files, zizmor clean at the new scope both offline and with the online audits CI runs.


vyncint (changes requested):
The substance here is right and well argued. I reproduced both halves:

```console
$ zizmor --persona=pedantic --no-online-audits .github/
completed .github/actions/report/action.yml
completed .github/dependabot.yml
No findings to report. Good job! (18 ignored, 1 suppressed)
```

and, with an interpolation planted in the action, the wider scope catches it while the old scope stays green — which is #389 in one experiment:

```
error[template-injection]: code injection via template expansion
  --> .github/actions/report/action.yml:50:26
# same tree, old scope: "No findings to report. Good job!"
```

The judgement call you made is the hardest part of the change and you got it right: you checked the wider scope for new findings *before* claiming it was clean, found the `dependabot.yml` one the workflow-only glob had never parsed, and wrote the acceptance in `.github/zizmor.yml` with a rationale saying why fixing it belongs in a different pull request. Recording the concrete numbers there — GitHub's 3-day default against zizmor's 7 — means whoever revisits it does not have to re-derive them.

**One thing to fix before merge.** You updated the inline comment in CONTRIBUTING §1 to `# workflow and action audit`, but the two matching places in `.github/workflows/ci.yml` still say workflows only:

- line 328: `- name: Audit workflows at pedantic level`
- line 319: `# Workflow security audit (template injection, credential persistence,`

Both now understate what runs, and the step name is the string that shows in the Checks UI and in the job log, so it is the one a reader will trust. Something like `- name: Audit .github at pedantic level` and a matching tweak to the comment on 319 would do it. Since you already fixed the wording in CONTRIBUTING, I read this as an oversight rather than a decision — otherwise the change is ready.

Ping me when that is in and I will merge.

maskjelly:
Both places are updated: the step name is now `Audit .github at pedantic level`, and the job comment names the composite action the repository ships for other repositories to execute rather than the workflows alone. Rebased on main and force-pushed — the job's run line is unchanged, so the reproduced behaviour stands. Ready for your merge.

vyncint (approved):
Approved. The wider scope is the right call — `.github/actions/report/action.yml` is the one artefact this repository ships for *other* repositories to execute, and it was the one file the audit never read.

Two things I checked beyond CI, because #433 landed a stricter `gates-listed` after this branch was cut: merged against current main the gate still passes (19 cargo commands, 13 script gates), and the zizmor path now reads `.github/` identically in `ci.yml` and CONTRIBUTING §1, which is what the pin check holds together.

Accepting the two `dependabot-cooldown` findings rather than fixing them here is the right split — the cooldown is a dependency-update policy decision and does not belong in a PR about audit scope. The `.github/zizmor.yml` entry says so in the file, which is where someone will look.

Agreed on leaving the job's display name alone; renaming it touches the required-check configuration and is not worth coupling to this.

PR #439, opened 2026-09-17
test(cli): cover the four promised inspect flags

### LINKED ISSUE #367: Four `inspect` flags are part of the promised CLI contract and no test exercises any of them
**Today** — `docs/STABILITY.md` promises the CLI's "commands, flags, exit
codes and input formats", checked by `crates/termlens-cli/tests/cli.rs` and
`.github/scripts/check-cli-contract.sh`. Four flags appear in neither.
Counted on 0.11.1:

```console
$ for f in --env --inherit-env --idle --timeout; do
>   printf '%-14s cli.rs=%s contract=%s\n' "$f" \
>     "$(grep -c -- "\"$f\"" crates/termlens-cli/tests/cli.rs)" \
>     "$(grep -c -- "$f" .github/scripts/check-cli-contract.sh)"
> done
--env          cli.rs=0 contract=0
--inherit-env  cli.rs=0 contract=0
--idle         cli.rs=0 contract=0
--timeout      cli.rs=0 contract=0
```

All four work today — this is a coverage gap, not a breakage:

```console
$ MYVAR=zzz termlens inspect --size 60x3 --inherit-env sh -c 'echo "[$MYVAR]"' | sed -n 2p
[zzz]
$ MYVAR=zzz termlens inspect --size 60x3 sh -c 'echo "[$MYVAR]"' | sed -n 2p
[]
```

**Why it is worth fixing** — `--env` and `--inherit-env` exist so that
`inspect` shows the screen your *test* will see rather than the one your
shell produces, which is the whole point of
[#263](https://github.com/vyncint/termlens/issues/263). A regression in
either is silent: the screen still renders, it is simply the wrong one.
`--timeout` and `--idle` decide when the snapshot is taken, so a regression
there turns into someone else's flake.

**Fix** — add cases to `crates/termlens-cli/tests/cli.rs`, which already
drives the binary through a PTY with the library, and the matching one-line
checks to `.github/scripts/check-cli-contract.sh` beside the `--cwd` pair.
Suggested shapes: a variable set in the caller's environment is absent by
default, present with `--inherit-env`, and settable with `--env`; `--env`
splits on the first `=` so `--env A=b=c` sets `A` to `b=c`; a program that
outlives `--timeout` is snapshotted and reported as still running. Keep the
deadline generous on any wait that also covers the spawn — CONTRIBUTING §3
explains why.

**Done when**
- [ ] `--env`, `--inherit-env`, `--idle` and `--timeout` each have a case
      in `cli.rs`.
- [ ] `check-cli-contract.sh` covers `--env` and `--inherit-env`, so the
      published binary is checked too.
- [ ] The suite still passes with `cargo test --workspace --all-features`.

---

maskjelly:
Closes #367.

`docs/STABILITY.md` promises the CLI's commands and flags, checked by `cli.rs` and `check-cli-contract.sh` — and four `inspect` flags appeared in neither: `--env`, `--inherit-env`, `--idle`, `--timeout`. This is coverage, not a fix: all four work today. But `--env`/`--inherit-env` decide which screen a test sees (a regression still renders a screen, just the wrong one) and `--idle`/`--timeout` decide when the snapshot is taken (a regression becomes someone else's flake).

**Cases added to `cli.rs`:**

- the caller's environment is absent by default and present with `--inherit-env` (the variable is set on the termlens process itself);
- `--env` sets a variable, and `--env KEY=b=c` keeps everything after the first `=` (the shell reads `[b=c]`);
- `--idle` decides what a still-running child's snapshot holds: a short window excludes late output, a long one includes it;
- `--timeout` snapshots a child that outlives it and excludes output past the deadline — the exclusion is what catches a flag-ignored regression, since the longer default would have held it.

**Mutation-checked, as §3 asks:** four mutations — `--inherit-env` made a no-op, the `--env` value truncated at the second `=`, `--idle` hardcoded at the call site, `.timeout` hardcoded to the default — each turned its case red with the expected wrong screen (`[]` under `--inherit-env`, `[b]` for the split, `first` only for the long idle, `first second` for the deadline), and each was restored to green. Transcripts in the PR review notes if wanted.

**Contract script:** `--env` (malformed pair exits 2; sets a variable; the env is cleared by default) and `--inherit-env` (keeps the caller's) now sit beside the `--cwd` pair, so the published binary is held to them too.

**One honest gap:** `--env =b=c` (an empty key) is not asserted; the environ entry is `=b=c` either way and a shell ignores it, so the case pins less than it looks like it might. Left untested rather than pinned as intended behaviour.

**Stress note:** these add six PTY spawns across four tests, so CONTRIBUTING §3's stress trigger applies — I cannot dispatch it from a fork; could you run `stress.yml` on this branch?

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (cli.rs 20/20), `check-cli-contract.sh` PASS including the new lines. No CHANGELOG entry (test-only, matching the `test:`/`fix(test):` precedent).


vyncint (changes requested):
These are good tests, and I mutated the CLI six ways to check them. Five mutations are caught: making `--inherit-env` a no-op, dropping `--env` pairs, never clearing the environment, ignoring `--timeout` and ignoring `--idle` — most of them by both `cli.rs` and the contract script, which is exactly the redundancy worth having, since the script also holds the *published* binary to them.

Covering both sides of the `--idle` window in one test — 300 ms excluding the late output, 3000 ms including it — is what makes an ignored `--idle` impossible to miss, and my mutation confirms it fails that test and only that test.

**One assertion cannot fail, and it is the one the test is named after.** In `crates/termlens-cli/tests/cli.rs`, `inspect_env_sets_a_variable_and_splits_on_the_first_equals` asserts that `--env TERMLENS_INSPECT_SPLIT=b=c` yields `[b=c]`. I changed `split_once` to `rsplit_once` in `main.rs` — precisely the bug that assertion is for — and it still passed:

```console
$ cargo test -p termlens-cli --test cli inspect_env
test result: ok. 1 passed; 0 failed
```

The reason is that the pair is re-joined with `=` on its way into the child's environment block, so `("A=b", "c")` and `("A", "b=c")` both arrive as `A=b=c`, and the shell splits that on the first `=` for you. The child cannot tell the two apart. The first half of the test does bite; this half is the "gate that cannot fail" shape.

What does tell them apart is an **empty key**, because that is the one case where the two splits disagree about whether to refuse. I measured both:

```console
# correct, split_once
$ termlens inspect --size 40x3 --env '=a=b' /bin/echo hi
termlens: bad --env "=a=b", expected e.g. NO_COLOR=1     # exit 2

# rsplit_once
$ termlens inspect --size 40x3 --env '=a=b' /bin/echo hi
size: 40x3  cursor: 1,0
hi                                                        # exit 0
```

Could you keep the `[b=c]` case, which still documents the intent nicely, and add a `--env '=a=b'` case asserting exit 2 and the `bad --env` message? A matching line in `.github/scripts/check-cli-contract.sh` beside your malformed-pair check would hold the published binary to it too.

**Two smaller notes.** Your harness deadlines are right — 30 seconds on the wait that also covers the spawn, exactly as CONTRIBUTING §3 asks. But the CLI's own `--timeout 1` still has to cover the shell's exec and its first `printf`, and the test then asserts that output is on screen. I hammered it — 40 runs under 60 spinners on 12 cores — without a failure, and the Windows ignore keeps the loaded-runner case away from it, so I am not asking you to restructure. But the margin is about a second. Raising `--timeout` to 3 and the child's sleep to 4 would keep both discriminations and buy real slack.

And these four tests add seven PTY spawns, so §3 wants a `stress.yml` run on the branch. A fork cannot dispatch it, so ask in the body and I will.

maskjelly:
All three points addressed and force-pushed.

**The assertion that could not fail now has one that can:** `--env '=a=b'` is asserted to exit 2 with `bad --env "=a=b", expected e.g. NO_COLOR=1`, and the same case is a line in `check-cli-contract.sh` beside the malformed-pair check, so the published binary is held to the first-`=` parse too. I verified the two splits disagree exactly as your measurement shows before wiring it (`rsplit_once` passes `=a=b` through as a valid empty-key pair). The `[b=c]` case stays, with a comment saying why it cannot see the difference.

**Margins raised as you suggested:** `--timeout 3` and the child's sleep at 4 in both timing tests — and that surfaced a real race worth recording. With sleep 4 and `--idle 3000`, the silence window from `first` (≈0.03 s) comes due at ≈4.0 s, the same instant `second` is written, so whether `second` is on the deadline snapshot was decided by tens of milliseconds — it flaked immediately here, in both directions depending on load. The long case now uses `--idle 5000`, above the timeout, so the wait's own bound ends it after `second` has certainly arrived and before its silence could ever be satisfied; neither side races the 4 s mark. The child also carries a trailing `sleep` so its last write cannot be lost to the platform's write-then-exit PTY teardown (docs/DESIGN.md) — that bit the previous shape too. The file ran green three more times on a loaded machine after the change.

**Stress:** the four tests add the seven PTY spawns; a fork cannot dispatch `stress.yml` — could you run it on this branch? (The PR body asks the same.)

vyncint (approved):
Approved. Coverage of behaviour that decides which screen a test sees is worth as much as a fix, and the framing is right: a regression in `--env`/`--inherit-env` still renders a screen, just the wrong one, and a regression in `--idle`/`--timeout` becomes someone else's flake.

The mutation transcripts are the part that makes this reviewable, and the `--env =a=b` reasoning is the good kind of honesty — noting that the empty-key case is the only one where `split_once` and `rsplit_once` disagree about refusing, and that the `[b=c]` assertion alone would pass under both, is a sharper argument than the test it justifies.

Verified your five new `check-cli-contract.sh` lines against the **published** 0.11.1 binary, not only the tree: all five pass there, so unlike a contract assertion added beside a fix, these do not make the daily `install.yml` run any redder than it already is. That mattered today and I checked it specifically.

**Stress: 15/15.** Run on your branch merged with #437 onto current main, all three OSes, five thread counts — the six PTY spawns are fine.

PR #440, opened 2026-09-17
fix(ci): resolve link-gate relative links against the linking file

### LINKED ISSUE #384: Three links in `docs/LIMITATIONS.md` are 404s, and the link gate cannot see them: it resolves relative links against the repository root, not the linking file

maskjelly:
Closes #384.

**The bug, twice over.** `docs/LIMITATIONS.md` links its siblings as `docs/DESIGN.md` — right from the repository root, `docs/docs/DESIGN.md` from where the file sits — and the gate resolved every relative link against the root, so it called the file green while three links were dead. `ci.yml` also never passed the file to the gate at all.

**Three parts, as the issue asks:**

1. **Resolver**: a relative link now resolves against the directory of the file that contains it, the way GitHub/crates.io/browsers do, and the failure prints the path it resolved to (the clue the old diagnostic withheld).
2. **Selftest**: the good fixture's in-repo links now climb out of its own depth, so it passes *only* under the file-relative rule, and a new `contributing-root-relative/` fixture links paths that exist at the root and nowhere near the file — the exact #384 shape. 8 expectations.
3. **Links + invocation**: the three links are fixed (`[DESIGN.md](DESIGN.md)` etc.), and `docs/*.md skills/termlens/SKILL.md` join the invocation in `ci.yml` and CONTRIBUTING §1 — 53 targets across 10 files, all green.

**Evidence:** before, the gate on LIMITATIONS exits 0 (wrongly); with the resolver alone it fails on both distinct targets with the resolved-path diagnostic; after the link fix the wider invocation is green. Selftest with the gate reverted to HEAD fails both new cases; restored, 8/8.

**Not done:** the `docs/*.md` prose still writes `` `docs/DESIGN.md` `` in backticks where it is not a link — untouched, the gate cannot see it and it is correct as prose. No CHANGELOG (docs/CI, matching #352/#407 precedent). `check-ci-gates-listed.sh` still passes.


vyncint (approved):
Thanks! Merging.

I tested the thing that matters most for a gate fix: does the selftest fixture isolate the *new* rule, or does it go red for the old reason? Yours isolates it. I reverted only `check-readme-links.sh` to main, kept every fixture, and the new expectation flipped from correctly-failing to wrongly-passing:

```
FAIL  CONTRIBUTING.md: a link good from the root only fails (#384)  exit 0, expected nonzero
```

That is exactly what a gate selftest is for, and it is a trap a previous pull request in this repo fell into — a fixture that failed because its target did not exist at all, so reverting the logic changed nothing.

The extra error line is the best thing in the diff:

```
docs/LIMITATIONS.md LINK: "docs/DESIGN.md" is linked but not in the repository
  From this file it resolves to "…/docs/docs/DESIGN.md".
```

That turns a puzzling message into a self-explaining one. Whoever hits this next will not have to work out what the gate meant.

I also checked the rule that could most easily have been broken in passing: README stays absolute-only, because crates.io rewrites relative links against the crate directory. Appending a relative link to the real README under your code still fails with the right message, and the absolute-only branch never touches the new directory variable, so the new resolution cannot leak into it. All three LIMITATIONS links resolve, the widened invocation covers 53 targets across 10 files, and CONTRIBUTING §1 carries the byte-identical command so `gates-listed` keeps the two together.

Rewriting the `contributing-good` fixture's prose to explain why it now needs `../../../` was a good instinct — the fixture documents its own depth instead of looking like a mistake.

One optional follow-up, not worth holding this for: `tools/link-gate-selftest/run.sh` line 55, the "this repository's own files pass" case, still passes only three files and was not widened alongside ci.yml. Nothing is unguarded, since CI does check the docs, but the selftest no longer mirrors what CI runs.

PR #441, opened 2026-09-17
feat(tools): add a preflight runner for the CONTRIBUTING gates

### LINKED ISSUE #368: Running the local gates before a first PR means typing 33 commands by hand, and there is no one command that does it
**Today** — CONTRIBUTING §1 lists every CI gate so that "nothing in CI
should surprise you". It is 33 command lines, and the repository offers no
runner for them:

```console
$ awk '/^```sh$/{n++; next} /^```$/{next} n==2' CONTRIBUTING.md | grep -c '^[a-zA-Z.]'
33
$ ls | grep -iE 'justfile|makefile|xtask|Taskfile'
$ echo $?
1
```

So a contributor either pastes 33 lines, or pushes and learns from CI which
one they skipped — which is exactly the "avoidable bad first experience"
the section says it exists to prevent.

**Why it is worth fixing** — the list is long because the project really
does gate on that much, and that is not going to shrink. The cost lands
entirely on the newest contributor, who is least able to tell which of the
33 their change could possibly have broken. One command also makes the
list *testable* rather than aspirational.

**Fix** — add `tools/preflight.sh`: run each gate in order, print one
`ok`/`FAIL` line each in the style `tools/link-gate-selftest/run.sh`
already uses, keep going after a failure, and exit non-zero if any failed.
Two things make this smaller than it sounds:

- `.github/scripts/check-ci-gates-listed.sh` already extracts the command
  list out of CONTRIBUTING §1 with `awk`/`sed`. Reuse that extraction —
  factor it into a small function or a second script both call — so the
  runner cannot drift from the documented list the way a hand-copied array
  would.
- Skip, with a named message, the gates whose tooling may be absent
  (`cargo-deny`, `cargo-semver-checks`, `pipx`, the Windows target) rather
  than failing on them, and support something like `--fast` that runs only
  fmt, clippy and the test suite.

Then point at it from CONTRIBUTING §1 as the one-line version, keeping the
full list underneath — the list is also documentation of what each gate is.

**Done when**
- [ ] `tools/preflight.sh` runs the §1 gates, reports each, and exits
      non-zero if any failed.
- [ ] Its command list comes from CONTRIBUTING §1 rather than a second
      copy, so `gates-listed` keeps both honest.
- [ ] Absent optional tooling is skipped by name, not reported as failure.
- [ ] CONTRIBUTING §1 mentions it in a line or two.

---

maskjelly:
Closes #368.

**The runner:** `tools/preflight.sh` reads the gate list out of CONTRIBUTING §1 via a shared extractor, runs each command in order, prints one `  ok  /  FAIL  /  skip` line per gate in the link-selftest's style, keeps going after a failure, and exits non-zero if any failed. `--fast` narrows to fmt/clippy/test. A gate whose tooling is absent (cargo-deny, cargo-semver-checks, pipx, the 1.85 toolchain, the Windows target) is skipped by name with its install hint, never failed.

**No second source of truth:** the §1 extraction that `check-ci-gates-listed.sh` already performed moved into `.github/scripts/extract-gates.sh`, and both scripts call it. Proven equivalent: `check-ci-gates-listed.sh`'s output and its normalised command list are byte-identical before and after the refactor.

**Self-tested:** `tools/preflight-selftest/run.sh` (next to the link and report-pin selftests, and wired into the `gates-listed` job) runs fixtures where the first gate fails and the second passes — both must be reported, exit non-zero — and an all-passing fixture exits zero. It cannot run on a machine missing the optional tools, which is the point of the skip design.

**Real runs here:** `--fast` 318 s (11 ok, 1 skipped: the Windows target), full mode 176 s warm (23 ok, 8 skipped, 0 failed).

**Judgement calls, stated:** `tools/preflight.sh` itself is not listed inside §1's block (it would recurse); the pointer is prose above the block. `tools/preflight-selftest/run.sh` *is* listed, matching the three existing selftests, so a full preflight runs its own guard — happy to drop that line if you want the block frozen. Skip detection is keyed on documented command text, so a future optional-tool gate needs one `case` arm.


vyncint (approved):
Thanks! Merging — this is the strongest pull request of the round.

The selftest is why. I mutated `tools/preflight.sh` two independent ways and it caught both:

```
# the runner can never report failure
FAIL  a failing gate is reported and the run continues   exit 0, expected nonzero
# the runner stops at the first failure instead of continuing
FAIL  a failing gate is reported and the run continues   missing ^  ok    true
```

The second is what makes it a real selftest rather than an exit-status coincidence: your grep patterns assert the *keeps-going* property, not just the status. And it uses the same `expect`/`out.log`/padded-`ok` shape as the three existing gate selftests, so it reads as if it had always been there.

Factoring the extraction into `.github/scripts/extract-gates.sh` with two named callers, instead of copying the awk into the runner, is exactly what #368 asked for — the runner cannot drift from the documented list because it *is* the documented list. I diffed the refactor's behaviour and it is behaviour-preserving: `check-ci-gates-listed.sh` still reports the same count and passes. The header comment explaining why the `RUSTDOCFLAGS` and `+toolchain` normalisation stayed behind, so preflight can run each line verbatim, is the kind of reasoning that stops the next person undoing it.

I ran the whole thing on a clean checkout: 28 ok, 3 skipped, 0 failed, about six minutes. Every skip category behaves as promised — cargo-deny, cargo-semver-checks, pipx, the 1.85 toolchain and the Windows target are each skipped *by name* with the install command, never failed. `--fast` narrows correctly by matching the documented command, so a new fmt, clippy or test line in §1 is picked up without editing the script, and `--bogus` exits 2 as the header says.

Two observations, neither worth changing now. The run prints about 5,600 lines for 31 verdicts, where `tools/link-gate-selftest/run.sh` captures output and shows it only on failure — I think passing it through is defensible for a six-minute run, since five minutes of silence is worse, but it is a reasonable follow-up if you disagree. And `FAIL` lines go to stderr while the closing failed-list goes to stdout; nothing is lost, the streams are just inconsistent.

Note that #433, which widens `gates-listed` to see script gates, is still open pending a small fix. Once it lands, the success line will tick to `13 script gates` and your new selftest gate will be covered by it. The two merge cleanly in either direction — I checked.

PR #442, opened 2026-09-17
docs: send usage questions to Discussions Q&A

### LINKED ISSUE #370: Asking a usage question means filling in a bug report's four required fields, while the Discussions tab sits unlinked
**Today** — `.github/ISSUE_TEMPLATE/config.yml` turns blank issues off and
sends "Question / usage help" to the bug-report form with a `[question]:`
title:

```yaml
blank_issues_enabled: false
contact_links:
  - name: Question / usage help
    url: https://github.com/vyncint/termlens/issues/new?labels=question&template=bug_report.yml&title=%5Bquestion%5D%3A+
    about: For usage questions, open a bug report and delete the sections that don't apply.
```

The advice to "delete the sections that don't apply" cannot be followed:
four of that form's fields are `required: true`, and GitHub refuses to
submit without them.

```console
$ grep -B6 'required: true' .github/ISSUE_TEMPLATE/bug_report.yml | grep 'label:'
      label: termlens version
      label: Operating system
      label: Minimal reproduction
      label: Expected vs actual
```

Meanwhile Discussions is already enabled on the repository and nothing
points at it:

```console
$ gh api repos/vyncint/termlens --jq .has_discussions
true
$ grep -ri discussion .github/ISSUE_TEMPLATE/ README.md CONTRIBUTING.md | wc -l
       0
```

**Why it is worth fixing** — "how do I wait for this?" is the first thing
a new user of a test harness has, and the project's answer is a form
demanding a minimal reproduction of a bug they are not reporting. Some
will write a fake bug; most will write nothing, and their question was the
cheapest signal the project could have had about which part of the API
reads badly.

**Fix** — in `config.yml`, point the question link at
`https://github.com/vyncint/termlens/discussions/new?category=q-a` (create
the Q&A category first if it is not there) and reword `about` to match.
Either way the wording must stop telling people to delete required fields.
If Discussions is not the wanted route, the alternative is a third
template with no required field beyond the question itself — say in the PR
which you chose.

**Done when**
- [ ] The question route leads somewhere a question can actually be posted
      without inventing a reproduction.
- [ ] No template text instructs deleting a field GitHub requires.
- [ ] README or CONTRIBUTING mentions where questions go, in a line.

---

maskjelly:
Closes #370.

The "Question / usage help" contact link sent people to the bug-report form, whose four required fields cannot be deleted — so "delete the sections that don't apply" was impossible advice, and asking a usage question started with a version number, an OS and a reproduction before GitHub would accept it. Discussions is enabled, has a Q&A category, and nothing pointed at it.

**Shape chosen:** the contact link now opens a Q&A discussion (`/discussions/new?category=q-a`), the advice says what is true (a reproduction is not needed for a question), and the README's getting-help sentence names it beside the security-report pointer.

**Evidence:** `has_discussions` is true; the GraphQL category listing confirms the `q-a` slug (`Q&A`, answerable); the URL form is the one GitHub itself uses for Discussions contact links. Honest limit: an anonymous HEAD request cannot distinguish a valid slug from an invalid one (both redirect to login), so the slug is proven by the API listing, not by curl. YAML parses and matches the issue-template config schema.

**Not done:** the bug form's required fields are untouched — a real bug report should carry them, and questioners no longer reach that form. `docs/HANDOFF.md` still says "Discussions stayed off" in one paragraph; that is a historical phase note and a separate edit if you want it. No CHANGELOG (template/docs wiring, matching #407 precedent).


vyncint (approved):
Thanks! Merging.

I checked the thing most likely to be wrong here rather than assuming it: the category really exists, so the link will not 404.

```console
$ gh api graphql -f query='{repository(owner:"vyncint",name:"termlens"){discussionCategories(first:25){nodes{name slug isAnswerable}}}}'
{"name":"Q&A","slug":"q-a","isAnswerable":true}
```

The slug in your URL is the slug GitHub has. Worth knowing for next time: hitting `/discussions/new` while logged out redirects to the login page for *any* category name, so that probe proves nothing on its own — the category listing is the evidence.

Two things I liked. You rewrote the `about:` text rather than only swapping the URL, so "a bug reproduction is not needed" answers the exact hesitation the issue described, in the place the person reads it. And you landed the README signpost in the same pull request, so the route and the pointer to it ship together instead of leaving a half-done state on main.

Leaving `blank_issues_enabled` at `false` is the right non-change. It was false because there was no other route; now that there is a working one, turning it on would reopen the write-a-fake-bug path the issue is about.

Issue #340, opened 2026-09-10
termlens inspect prints a screen that termlens render and diff refuse to read


vyncint:
**Today** — `termlens inspect` is the only way the CLI produces a screen, and `termlens render` and `termlens diff` are the only commands that consume one. Its output cannot be fed to either.

```console
$ termlens inspect sh -c 'printf hi' > screen.txt
$ termlens render --text screen.txt
termlens: screen.txt: could not parse a saved screen: line 26: expected `styles:` or the end of the text after the 24-row grid, got "--- exited: exit code 0 ---"
$ echo $?
2
```

Measured against **published 0.10.2** (`cargo install termlens-cli --version =0.10.2`), no path dependency. `diff` fails identically.

The trailer is written to **stdout** ([main.rs:420](https://github.com/vyncint/termlens/blob/3ff0922d8cc99bf610f291c34ee47262d667485e/crates/termlens-cli/src/main.rs#L420), and [:428](https://github.com/vyncint/termlens/blob/3ff0922d8cc99bf610f291c34ee47262d667485e/crates/termlens-cli/src/main.rs#L428) for the deadline case), so a redirect captures it. There is no flag to suppress it — `inspect --help` offers `--size`, `--timeout`, `--idle`, `--inherit-env`, `--ansi`, `--env`, and nothing else — `--ansi` emits it too, and `TERMLENS_ARTIFACT_DIR` is not honoured by `inspect`, so **no CLI route produces a file the CLI will read back**. The workaround is `sed '/^--- exited:/d'`, which a user has to invent.

## Why this is a defect and not just a missing convenience

The usage text's own lead sentence invites exactly this ([main.rs:32](https://github.com/vyncint/termlens/blob/3ff0922d8cc99bf610f291c34ee47262d667485e/crates/termlens-cli/src/main.rs#L32)):

> A saved screen is the snapshot text format **termlens prints** — an insta `.snap` with or without its header, the block a wait error prints, a `TERMLENS_ARTIFACT_DIR` file — or the JSON…

The enumeration that follows does not include `inspect`'s output, but "the snapshot text format termlens prints" is read as covering the thing `termlens` most visibly prints.

And the parser is **already asymmetric about decoration** — it tolerates a leading insta header and rejects anything trailing:

| input | `render --text` |
|---|---|
| screen with a leading `---`/`source:`/`---` header | exit **0** |
| screen with the `--- exited: …` trailer | exit **2** |
| screen with any other trailing line | exit **2** |

[...]

PR #443, opened 2026-09-17
fix(example): align the inspect example with the CLI contract

### LINKED ISSUE #372: The `inspect` example and `termlens inspect` are two implementations of one tool, and they now disagree on the exit code and on which stream the trailer uses

maskjelly:
Closes #372.

**Align, not retire** (option 1): the example is the one worked example of driving `Terminal` from a `main`, and `docs/HANDOFF.md:255` points readers at `cargo run --example inspect`. Retiring it would delete a documented entry point and its 289-line test.

**Aligned to the promised contract:** own-failure paths exit 2 with the CLI's diagnostic shape (program-name prefix the only difference), the trailer goes to stderr so stdout stays a saved screen (the #340 shape), and the module docs teach the new contract instead of the old one. It also gains `--cwd`, copied from the CLI's parser — same `take`, same `is_dir` check, same diagnostic — and its `--version` prints the CLI's exact line. (One correction to the issue text: `--version` existed since #243 with a divergent shape; it is aligned, not added.)

**Tests:** the example's test updates with it — the changed assertions pin the new contract, timeout and relative-path cases assert the stderr trailer, and `--cwd`/`--version` cases are new. With the example reverted the new assertions fail four ways (`left Some(1) right Some(2)`, the old version line, missing stderr trailers); with the change 6/6 pass. Manual parity checked against the CLI: spawn failure exits 2 with the same message shape, `--version` byte-identical, and the CLI's `render`/`diff` read the example's stdout back.

**Not done:** `docs/HANDOFF.md` unchanged — its wording ("shows you any program through termlens's eyes") is still true. CHANGELOG carries a short `### Fixed` bullet, matching the precedent for example changes.


vyncint (approved):
Thanks! Merging.

You picked option 1, alignment, and said so explicitly rather than leaving me to infer it. I built both binaries and drove them side by side across an eighteen-case matrix: every failure of the tool itself now exits 2 in both — no arguments, bad `--size`, a missing flag value, an unknown option, a spawn failure, a bad `--cwd`, a malformed `--env`, an empty-key `--env`. And for a plain run, an `--ansi` run and a timeout snapshot, the example's stdout and stderr are byte-identical to the CLI's, trailing newline included.

The best thing in the diff is in the test, not the example: `Screen::parse(&stdout).expect("stdout is a saved screen")`. That does not merely check the trailer moved off stdout — it proves what remains is a file the tool reads back, which is the actual point of #340. Checking the property rather than the symptom is what makes that test worth keeping.

I also liked that the body corrects the issue rather than quietly diverging from it: `--version` existed already with a divergent shape, so it is aligned, not added. Saying so is more useful than silently doing the right thing.

Reverting the example fails four of the six tests. Leaving `docs/HANDOFF.md` alone was right, since the example survived, and it teaches no contract.

The only divergences left are cosmetic: the program name in the usage banner and the hint, and two extra sentences in the CLI's help.

One thing I am watching rather than asking you to fix: #437 adds `--flag=value` to the CLI. Today you and it agree that `--size=60x3` is an unknown option; once that merges, the CLI will accept it and the example will not — a fresh divergence on the very axis you just closed. I have asked that PR to carry its parser into the example, and merging yours first is what makes that the cheaper order.

PR #444, opened 2026-09-17
fix(emu): report OSC 4/10/11 colour sets as unsupported

### LINKED ISSUE #392: Setting a palette or a foreground colour with `OSC 4`, `OSC 10` or `OSC 11` is neither applied nor listed by `unsupported()`
**Today** — the unhandled-sequence filter exempts these three OSC numbers
by number, but only their *query* forms are answered. A set is dropped by
both halves and reported by neither:

```
^[]4;1;rgb:ff/00/00   -> unsupported []
^[]10;#ff0000         -> unsupported []
^[]11;#00ff00         -> unsupported []
^[]11;?               -> unsupported []      (correctly: it is answered)
^[]777;x              -> unsupported ["^[]777;x"]
```

**Why it is worth fixing** — `docs/LIMITATIONS.md` promises that
`Screen::unsupported()` "lists every sequence the emulator did not
implement", and the module's own header says the list means the sequences
*termlens* did not honour. The query handler already draws exactly this
distinction in a comment — the `?` matters, because both codes also set,
and a set is not a question — and the filter contradicts it. An application
that repaints its palette gets a grid drawn in the wrong colours with
nothing anywhere saying so, which is the "plausible-looking wrong grid" the
accessor was added for.

**Fix** — in `crates/termlens/src/emu/unhandled.rs`, narrow the exemption
to the forms the responder covers: require a trailing `?` for 4, 10 and 11,
and leave 8 and 52 exempt as they are.

**Done when**
- [ ] `^[]11;#00ff00` appears in `unsupported()`; `^[]11;?` still does not.
- [ ] The existing "an application using only what is implemented reports
      nothing" test still passes — it sends the query form.
- [ ] A new case in `crates/termlens/tests/unsupported.rs` pins the
      set-versus-query split.

---

YuEfSaEDU:
The unhandled-sequence record exempted `OSC 4`, `OSC 10` and `OSC 11` by number, but only their `?` queries are handled: `10;?` and `11;?` are answered by the responder and `4;n;?` is named in a timeout. The same codes also set a colour, and a set is not a question — it is honoured by nobody, yet the exemption kept it out of the record, so an application repainting its palette got the wrong colours with nothing saying why (#392).

The exemption is now narrowed to the trailing-`?` forms; `OSC 8` and `OSC 52` stay exempt in every form. Applying the colours is deliberately out of scope: until the emulator honours a set, naming it in `unsupported()` is the documented contract, and honouring it is an emulator change rather than a filter fix. The PTY-level case runs on the Unix CI legs only, so the same set-versus-query split is pinned by a table-driven unit test that runs on every platform.

Closes #392.


vyncint (approved):
Thanks, and welcome — this is a strong first contribution.

You found the right seam. The exemption was keyed on the OSC *number*, and the fix keys it on the *shape* instead, which is what makes it a rule rather than a patch. Better still, the rule you chose is the one `seq.rs` already applies: its own comment says the `?` matters because both codes also set, and a set is not a question. Reading the responder and matching it, rather than inventing a second rule, is exactly right.

Three things I checked and liked:

Pinning the split at two levels. The PTY test is the honest one, but it is Unix-only, so the table-driven unit test carries the same distinction onto every platform. That is the reason the Windows leg is not just an ignored blank.

`OSC 8` and `OSC 52` stay exempt in **every** form, not only as queries — correct, because termlens really does honour those: a link becomes a span and a clipboard set becomes `Screen::clipboard()`. Getting that asymmetry right is the part most people would have flattened.

Saying plainly that applying the colours is out of scope. Until the emulator honours a set, naming it is the documented contract, and you kept the change to the filter.

I reverted your filter and both new tests failed, the unit one and the PTY one. That is what I want from a test like this.

**I pushed one commit, `f181daa`,** for a narrow hole in the new rule. Keying on a trailing `?` anywhere in the parameters is right for `OSC 4` — `seq.rs` names any `4;…?`, so filter and responder agree, and I added a mixed `4;1;rgb:…;2;?` row to show it. But `10` and `11` are answered only in their exact two-operand form, and xterm also reads `OSC 10;#ff0000;?` as set-then-query. Nothing answers that one, so exempting it dropped a sequence nobody honoured — the same hole as #392, one corner further in. The commit requires the two-operand form for 10 and 11, keeps any `4;…?`, and notes in the comment that the rule belongs to `seq.rs`. Two rows pin both edges and reverting either half fails them.

Two process notes for next time, neither a problem here. Your body said `(#392)` rather than `Closes #392`, so GitHub had not linked the issue — I edited it in so the issue closes on merge. And CONTRIBUTING §3 asks for `stress.yml` on a change that adds PTY spawns; a fork cannot dispatch it, so say so in the body and a maintainer will run it. I did: 15 legs green across Ubuntu, macOS and Windows at every thread count.

PR #446, opened 2026-09-17
fix(emu): keep shadow parser aligned

### LINKED ISSUE #390: An SGR the shadow parser drops entirely is rewritten to zero bytes, so a preceding ESC disappears and the attribute shadow diverges from the grid
**Today** — the shadow rewrites each SGR down to the attributes it tracks.
When it tracks none of them — `^[[31m`, `^[[7m`, `^[[38;5;1m` — the
rewrite is the empty string. The ESC that introduced the sequence goes with
it. If that ESC was also doing structural work in the primary stream,
terminating an unterminated OSC or aborting a half-written CSI, the shadow
stream loses that work and its parser then swallows the next printable
character as a final byte.

An application that sets a title without a terminator and then colours some
text:

```console
$ # payload: ESC]0;title ESC[31m A ESC[5m B ESC[0m C
thread 'main' panicked at crates/termlens/src/emu/vt100.rs:410:
assertion `left == right` failed: the attribute shadow diverged from the primary grid
  left: "ABC"
 right: "BC"
```

An ESC restarting a partial CSI does the same:

```console
$ # payload: ESC[1;2 ESC[31m A
assertion `left == right` failed: the attribute shadow diverged from the primary grid
  left: "A"
 right: ""
```

Both reproduce every time. In a release build there is no assertion, and
the result is silently wrong instead: blink lands one cell left of where
the application put it.

**Why it is worth fixing** — blink, conceal and strikethrough are read from
the shadow's cells, and the masked-password assertion the module exists for
is answered from them. So in release the answer comes from the wrong
column, and in debug — which is what every downstream `cargo test` runs —
the library panics on output a program legitimately produced. The module
header argues the two streams correspond "rather than leaving it to
argument"; this is a hole in that argument.

**Fix** — in `crates/termlens/src/emu/shadow.rs`, have the SGR rewrite emit
a complete sequence rather than nothing when it keeps no parameter, so the
ESC survives and the two parsers stay in step. `ESC[39m` is one candidate:
it is a well-formed SGR, and the shadow's cells are never read for
foreground colour. Whatever you choose, say in the PR why it cannot affect
the three attributes the shadow is read for.

This is a small change in a deliberately delicate module — `docs/DESIGN.md`
and `docs/STABILITY.md` §2 explain why the second parser is there. Read
both first, and open a draft PR early.

**Done when**
- [ ] Both payloads above snapshot without tripping the debug assertion.
- [ ] A unit test in `shadow.rs` asserts the rewrite of a fully-dropped SGR
      is non-empty and begins with `ESC`.
- [ ] A case in `crates/termlens/tests/styles.rs` pins blink landing on the
      cell the application set it on across an unterminated OSC.
- [ ] The stress workflow passes on the branch — this touches the emulator
      path every test drives.

---

Rayan-and-beyond:
Closes #390.

When the shadow drops every parameter from an SGR, it now emits a harmless foreground reset instead of an empty rewrite. That preserves the ESC/parser boundary while leaving blink, conceal, and strikethrough untouched, which are the only shadow attributes read back.

Validation:
- shadow unit tests: 10/10
- styles integration tests: 8/8
- focused Clippy and fmt
- `git diff --check`
- DCO/no-attribution checks


vyncint:
Closing in favour of #449, which landed the same fix for #390 — thank you @Rayan-and-beyond for diagnosing the dropped ESC independently; the two PRs agreed on the cause and the 0.11.2 changelog entry credits the fix. Your diagnosis is what made a second reading possible.

PR #447, opened 2026-09-17
feat(cli): add --json to render

### LINKED ISSUE #373: `termlens render` reads the JSON screen format and cannot write it, so there is no CLI route to a promised artifact
**Today** — the JSON a `Screen` serialises to is a promised, format-numbered
artifact: `docs/STABILITY.md` lists "the JSON shape (`serde` feature),
format `1`" beside the text format, and the CLI's `load()` accepts it as an
input. It cannot produce it:

```console
$ termlens render --help | head -1
usage: termlens render (--svg | --html | --ansi | --text) [--out PATH] <a>
$ termlens render --json a.snap
termlens: unknown option "--json" (try `termlens render --help`)
```

Everything needed is already linked in:

```console
$ grep -n 'serde' crates/termlens-cli/Cargo.toml
termlens = { version = "0.11.1", path = "../termlens", default-features = false, features = ["serde"] }
serde_json.workspace = true
```

**Why it is worth fixing** — the two formats are meant to be
interchangeable ways to save one screen, and the tool converts in one
direction only. Anyone wanting the JSON — to diff it in a pipeline, to feed
a dashboard, to attach a machine-readable screen to a bug — has to write a
Rust program to get what `termlens inspect app > screen.snap` already
captured. It is also the natural way to *check* the promise: a corpus file
should be able to make its own JSON twin through the CLI.

**Fix** — in `crates/termlens-cli/src/main.rs`, add `--json` to `render`'s
format list, serialising with `serde_json` — pretty-printed, since the
other renderings are for humans too, and ending with a newline like the
text one. `--out` and `-` should work with it unchanged. Update
`RENDER_USAGE` and the `render` line in the README's "Command line and CI"
section.

**Done when**
- [ ] `termlens render --json a.snap` writes the format-1 document, and
      `termlens render --text` reads that document back to the same screen.
- [ ] `--out` and stdin work with `--json`.
- [ ] `crates/termlens-cli/tests/cli.rs` covers the round trip, and
      `.github/scripts/check-cli-contract.sh` gains `--json` beside the
      other four formats.

---

maskjelly:
Closes #373.

`docs/STABILITY.md` promises the screen's JSON as format 1, and `render` and `diff` read it — but nothing could write it. Converting a saved text screen took a Rust program, so the two promised saved-screen formats were interchangeable in one direction only.

**The flag:** `termlens render --json` writes the document pretty-printed and newline-terminated — byte for byte the shape the compatibility corpus commits, so it diffs readably and `load` reads it straight back. `--out` and `-` behave as they do for the other four formats, and the usage lines (including the top-level help) name the new format.

**Evidence, all through the built CLI:** `render --json before.snap` twice is byte-identical (no timestamps anywhere); `render --text` of the emitted JSON equals `render --text` of the original text screen; `render --json` of that JSON is byte-identical to the JSON; stdin (`render --json -`), `--out`, and JSON-as-input all round-trip; `diff` reads the JSON against the text with no difference. The contract script gains the `--json` exit code, a `"format": 1` check and a read-back through `render --text`, so the published binary is held to the interchangeability too.

**Tests:** one `cli.rs` test pins all of the above (deserialising the output as a `termlens::Screen` to prove shape 1, not just key presence). Removing `--json` from the parser makes it fail with `unknown option "--json"`; restored, green.

**Not done:** untouched `crates/termlens-cli/README.md`, the crate description, `skills/termlens/SKILL.md` and the STABILITY/DESIGN prose — all already describe JSON as an input format; say the word if any of them should also advertise the new output. No CHANGELOG omission: a short `### Added` bullet is in the branch.

**Gates:** `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --all-features -- -D warnings`, `cargo test --workspace --all-features` (48 suites, 0 failed), `check-cli-contract.sh` PASS, `cargo test -p termlens --all-features --test compat` (5 passed — the JSON format is promised; nothing in the corpus is affected by adding a writer).


vyncint (approved):
Approved. The asymmetry this closes is a real one: STABILITY promises the JSON as a saved-screen format and `load` has read it from the start, so the CLI could consume a format it could not produce.

Verified against a build of this branch merged onto current main, not just the branch: `render --json` emits `"format": 1`; `render --text` of the emitted JSON is byte-identical to `render --text` of the original `.snap`; `render --json` of that JSON is byte-identical to itself (idempotent); and two runs over the same input produce identical bytes, so nothing timestamped leaks in. Contract script PASS on the merged tree.

One consequence to flag, which is not yours to fix and is already in hand: the two assertions you added to `check-cli-contract.sh` are correct against the tree and will fail against the **published** 0.11.1, which has no `--json`. That script is also run daily by `install.yml` against the latest release, so this adds to a red run that #435 already started for the same structural reason. I am fixing the cause — pinning that job's checkout to the tag of the version it installs — in a follow-up, rather than asking anyone to hold assertions back until a release.

Leaving `crates/termlens-cli/README.md`, the skill and the STABILITY/DESIGN prose untouched is the right scope call; they describe JSON as an input and remain true.

PR #465, opened 2026-09-17
fix(cli): inspect resolves on exit or idle, whichever comes first

### LINKED ISSUE #374: `termlens inspect` always waits the whole `--timeout` on a program that does not exit, so the deadline is a fixed cost rather than a limit
**Today** — `inspect` calls `wait_exit()` with the full deadline first, and
only consults `--idle` once that deadline has *expired*. For a TUI — which
never exits on its own — the deadline is therefore always paid in full,
however quickly the application finished drawing. Measured on 0.11.1 with a
program that paints in under a millisecond and then sleeps:

```console
$ time termlens inspect --size 40x3 --idle 200 --timeout 6 sh -c 'printf "READY\n"; sleep 300' >/dev/null
        6.074 total
$ time termlens inspect --size 40x3 --idle 200 --timeout 1 sh -c 'printf "READY\n"; sleep 300' >/dev/null
        1.072 total
```

Six seconds of waiting for a screen that was complete at 1 ms. The
`--idle 200` never got a chance to end it early.

**Why it is worth fixing** — `--timeout` exists so that a slow-starting
application can be inspected at all, and raising it is the documented way
to do that. But raising it charges the same wait to every run, including
the fast ones, so the flag punishes the person who needs it. TUIs are the
main thing `inspect` is pointed at, and for them the deadline is never a
deadline. `--idle` is already the right answer and is simply asked too
late.

**Fix** — in `inspect` in `crates/termlens-cli/src/main.rs`, resolve on
whichever comes first: the program exits, or its output has been silent for
`--idle`, both bounded by `--timeout`. `Terminal` has the pieces —
`wait_idle_for(idle, timeout)` returns as soon as the silence window is
met — so the shape is to race the two rather than to sequence them, and to
keep the three existing trailers meaning what they mean today. Update
`INSPECT_USAGE`, which currently describes the sequential behaviour.

There is a real trade-off to state in the PR: a 300 ms silence window can
catch an application that paints in bursts wider than that, which is why
the flag is adjustable and why `INSPECT_USAGE` already warns about it.
Waiting the whole deadline is the conservative choice; the argument for
changing it is that the conservative choice is currently not optional.

The timing of these heuristics is explicitly *not* promised in
`docs/STABILITY.md`, and no exit code or format changes, so this is a
behaviour improvement rather than a break — mention it under **Changed** in
the CHANGELOG.

**Done when**
- [ ] A program that paints and then idles is snapshotted after the idle
      window, not after the deadline, and the timing shows it.
- [ ] A program that exits is still snapshotted at exit with the
      `--- exited: … ---` trailer, and one that is genuinely still
      producing output until the deadline still reports
      `--- still running at the deadline …`.
- [ ] `crates/termlens-cli/tests/cli.rs` pins the fast case with a
      generous ceiling rather than a tight one — see CONTRIBUTING §3 on
      where a short deadline belongs.

---

maskjelly:
Closes #374.

`inspect` waited the whole `--timeout` first and only consulted `--idle` once that had expired, so a TUI — the thing `inspect` is pointed at — always paid the full deadline: the issue's `--idle 200 --timeout 6` took 6.07 s for a screen complete at 1 ms.

The wait now races the two: `wait_idle_for(idle, timeout)` returns on the first of the child closing its output or the silence window elapsing, with the deadline applied once, and a zero-budget reap attributes the arm so the three trailers keep their meanings — exited (reaped first, so the screen still holds the final output), or still running at the deadline. `INSPECT_USAGE` no longer describes the sequential behaviour.

**The trade-off, as the issue asks it be stated:** a 300 ms silence window can catch an application that paints in bursts wider than that. That is why the flag is adjustable and why the usage text already warns about it; the argument for resolving early is that raising `--timeout` for a slow starter used to charge the same wait to every run.

**One existing test changes.** The four-flags test from #439 pinned the sequential behaviour's second case: a five-second idle window above a three-second deadline held `second`, which arrives four seconds in. Under the one bound it cannot, so that half now asserts the deadline ends the wait and `second` is absent, with a ceiling on the return the sequential wait could not make (it used to return around eight seconds). The first case (a window satisfied inside the deadline) is unchanged.

### Tests

- Three new tests, one per arm: the idle window ends the wait long before a long deadline (ceiling asserted); a quick exit is reported without waiting for the idle window; a child that talks until the deadline still reports still running.
- `cargo test --workspace --all-features`: 583 passed, 0 failed.
- `tools/preflight.sh`: 26 ok, 0 failed, 7 skipped (optional tooling not installed: cargo-deny, cargo-semver-checks, pipx, 1.85 toolchain, Windows target).

### Stress

The stress workflow is running on the same commit in my fork, since the branch is not an upstream ref: [run 35246229578](https://github.com/maskjelly/termlens/actions/runs/35246229578). Happy to have it re-run on the upstream branch if you prefer that provenance.


maskjelly:
A note on the first stress run (fork, commit d75a9bd): it caught the arm-attribution race I was worried about — iteration 3 of 25 at 16 threads, the trailer said "still running at the deadline" for a program that had exited (`inspect` on `sh -c 'printf hi'`). The EOF that ends the idle wait can beat the child's status becoming collectable by a scheduling hair, so the zero-budget reap lost the race under load.

Fixed in the pushed commit: the reaping probe now gets a short grace (500 ms, the same width the library gives its own post-reap drain), paid only by a child that is genuinely still running and only after the wait has ended. Plus a fourth test pinning the other side: a child that closes its terminal but keeps running must still say "still running". Stress is running again on the new commit.

maskjelly:
Stress on the fixed commit is green: [run 35247718220](https://github.com/maskjelly/termlens/actions/runs/35247718220) — 15/15 shards passed (100 iterations split across them, all three OSes).

vyncint (approved):
The core of this is right, and the reasoning in the comments is the part I'd have had to reconstruct otherwise — particularly that `wait_idle_for` *is* the race rather than half of it, because it also returns on EOF. I confirmed that against the wait engine (`if guard.eof { return Ok(()) }`, ahead of the quiet check), which is what makes the reap probe sound. Measured on your branch: 6.07s → 0.76s for the TUI, 0.006s for a program that exits with `--idle 30000`, and 2.06s for a chatterer under a 2s deadline.

I pushed one commit on top rather than sending it back, since three things needed settling and they were quicker to do than to describe:

**The trailer.** `Ok` from `wait_idle_for` means quiet *or* EOF — neither is the deadline — so a TUI snapshotted 200ms into a 30s budget was being told `still running at the deadline`, the one thing that had not happened. Two trailers now, and `strip_inspect_trailer` matches `--- still running` as a prefix so both forms and every file written up to 0.11.2 still strip. Verified all four shapes round-trip through `render --text`.

**The example.** `examples/inspect.rs` mirrors the command by design (#372) and still had the sequential wait, documented as such.

**A test that had stopped testing.** `inspect_timeout_snapshots_a_program_that_outlives_it` uses the default 300ms window against a child that goes quiet immediately, so after your change the wait ended at 300ms whatever `--timeout` said — it would have passed against a deadline that did nothing. `--idle 5000` puts the window above the deadline so the deadline is measured again. Worth knowing as a general hazard of this change: any test whose child goes quiet early is no longer testing the deadline.

Also moved the CHANGELOG entry out of the \[0.11.2\] section — 0.11.2 published while this was open, and the rebase put your bullet next to the #390 one it was written beside, which had moved into the released section.

Thank you — this is the fix #374 wanted.

PR #466, opened 2026-09-17
docs: add XTVERSION and OSC 4 palette query to unanswered queries

### LINKED ISSUE #461: LIMITATIONS' list of deliberately unanswered queries omits XTVERSION and the `OSC 4` palette query
**Today** — `docs/LIMITATIONS.md` §"Waits, frames and queries" lists the
questions termlens deliberately leaves unanswered:

> Some questions stay deliberately unanswered — kitty's `CSI ? u`, DECRQSS,
> DA3, `OSC 12`, `OSC 52` *reads*, and the non-pixel `CSI … t` reports —
> because a guessed reply is worse than none. An application blocked on one
> is **named in the next timeout** rather than left to hang unexplained.

Two members of that set are missing from the list: **XTVERSION** (`CSI > q`)
and the **`OSC 4` palette query** (`OSC 4 ; n ; ?`). Both are recognized,
neither is answered, and both are named in a timeout —
`crates/termlens/tests/queries.rs::decrqss_and_palette_queries_are_named`
pins all three together:

```rust
("DECRQSS",   r"\eP$qm\e\\", "^[P$qm"),
("OSC 4",     r"\e]4;1;?\a", "^[]4;1;?"),
("XTVERSION", r"\e[>q",      "^[[>q"),
```

DECRQSS made the doc; the other two did not. XTVERSION arrived with
[#403](https://github.com/vyncint/termlens/issues/403) in 0.11.0 — the
CHANGELOG records it — and `docs/LIMITATIONS.md` was not updated with it.
`XTVERSION` appears nowhere in any `.md` file except that CHANGELOG line.

**Why it is worth fixing** — this list is the user-facing answer to "my
application hangs on startup, is that termlens?". Someone whose app probes
for terminal version (a common capability check — it is how a program asks
whether it is talking to kitty, WezTerm or something else) will look here,
not find XTVERSION, and conclude the hang is something else. The list's
value is being exhaustive; two gaps in six entries is most of that value.

**Fix** — add both to the sentence in `docs/LIMITATIONS.md`. Nothing else
changes: the behaviour is already correct and already tested, this is the
doc catching up.

While you are there, check the same list against
`crates/termlens/src/emu/seq.rs` for anything else recognized-but-unanswered
that never reached this file — that is how these two got missed, and a
third would be worth finding now rather than after the next report.

**Done when**
- [ ] The list names XTVERSION and the `OSC 4` palette query.
- [ ] `.github/scripts/check-readme-links.sh … docs/*.md` still passes.

---

Voyagerroc-Lab:
Resolves #461

### Summary
Adds XTVERSION and the \OSC 4\ palette query to the list of deliberately unanswered queries in \docs/LIMITATIONS.md\. Both sequences are recognized by the emulator but left unanswered (named in timeouts as context/cause), catching up the documentation with the existing behavior.


vyncint (approved):
Correct, and it lands exactly where the list was thin. I verified both additions are recognized-but-unanswered rather than merely undocumented: `queries.rs::decrqss_and_palette_queries_are_named` pins DECRQSS, `OSC 4` and XTVERSION together, and only DECRQSS had made this file.

Your branch is based on `ca4d73e`, before #384 repointed the DESIGN link in this file, so the diff shows `[docs/DESIGN.md]` as context. I trial-merged it: the three-way merge keeps main's `[DESIGN.md](DESIGN.md)` and applies your sentence, and the link gate passes on the result. Nothing to do.

Thank you.

PR #467, opened 2026-09-17
fix(cli): diagnose a missing --color value, not the raw Option

### LINKED ISSUE #452: `termlens diff --color` with no value swallows the next operand and reports it back as `got Some("before.snap")`
**Today** — `diff --color` with no value prints a Rust `Option`'s `Debug`
straight at the user, and quietly eats the next operand on the way:

```console
$ termlens diff --color before.snap after.snap
termlens: --color takes auto, always or never, got Some("before.snap")

$ termlens diff before.snap after.snap --color
termlens: --color takes auto, always or never, got None
```

`Some("before.snap")` and `None` are internals. Neither says the actual
problem, which is that `--color` needs a word after it and took a file
instead.

**Why it is worth fixing** — the `=` spelling of the same flag already has
the right shape, so the tool disagrees with itself about how one flag
reports one mistake:

```console
$ termlens diff --color=pink a b
termlens: --color takes auto, always or never, got "pink"
```

And `--color` is the flag most likely to be typed by hand at a prompt,
where forgetting the value is the easy slip.

**Fix** — in `diff` in `crates/termlens-cli/src/main.rs`, the arm matches
on `args.next().map(String::as_str)` and formats the whole `Option`. Take
the value out first, then:

- no value at all → the message every other flag uses, whose shape lives in
  `take()` a little further down the file: `` "{flag} needs a {kind}
  argument" `` — so, `--color needs a WHEN argument`.
- a value that is not one of the three → quote just the word, as the `=`
  arm does.

Worth a look while you are in there: the two arms (`--color` and
`--color=`) now spell the same list of three words twice, and the fix is a
natural moment to have them share one.

**Done when**
- [ ] `diff --color before.snap after.snap` exits 2 and says `--color`
      needs a value, without printing `Some(...)`.
- [ ] `diff before.snap after.snap --color` says the same thing, without
      printing `None`.
- [ ] `diff --color=pink a b` still names `"pink"`, and the three good
      values still work in both spellings.
- [ ] `crates/termlens-cli/tests/cli.rs` pins the missing-value message.

---

YuEfSaEDU:
The `--color` arm matched on `args.next().map(String::as_str)` and formatted the whole `Option`, so a missing value swallowed the next operand and reported it as an unknown color word:

    $ termlens diff --color before.snap after.snap
    termlens: --color takes auto, always or never, got "Some(\"before.snap\")"

Now the arm requires a word (the `take()`-shaped error used by sibling options):

    $ termlens diff --color before.snap after.snap
    termlens: --color takes auto, always or never, got "before.snap"
    $ termlens diff before.snap after.snap --color
    termlens: --color needs a WHEN argument

Both spellings (`--color WHEN` and `--color=WHEN`) share a single `color_when` helper holding the one list of WHEN words, so they cannot drift apart again.

Tests: `diff_names_a_missing_or_unknown_color_value` pins all three failure shapes (operand-swallow, flag-at-end, unknown `--color=pink`) and was verified to fail against the pre-fix code; `diff_accepts_the_three_color_words_in_both_spellings` covers `auto`/`always`/`never` in both spellings. `cargo fmt` clean, `cargo clippy -p termlens-cli --all-targets -- -D warnings` clean, `cargo test -p termlens-cli` 19 passed / 9 PTY-ignored.

Fixes #452


vyncint (approved):
This is more than the issue asked for and better for it. Sharing one list of the three words between the two spellings is the right shape — the duplication was the reason the two diagnostics could drift in the first place — and `color_when` returning the message rather than printing it keeps `fail` as the single exit.

Verified each path against your branch:

```
diff --color before.snap after.snap  -> --color takes auto, always or never, got "…/before.snap"
diff before.snap after.snap --color  -> --color needs a WHEN argument
diff --color=pink …                  -> --color takes auto, always or never, got "pink"
diff --color= …                      -> --color takes auto, always or never, got ""
```

`Some(` appears nowhere, and all six good spellings still exit 1 on differing screens. I especially like that the first test pins the shape at the front and the operand at the back rather than the whole string — that is the right call for a path `data()` Debug-escapes on Windows.

Thank you.

Issue #313, opened 2026-09-08
termlens render can only write to stdout, so an SVG has to be shell-redirected


vyncint:
**Today** — `termlens render --svg failing.snap` writes to stdout, and that is the only destination. Measured on 0.10.1: `render()` in `crates/termlens-cli/src/main.rs` ends in `print(&out)` with no path handling.

**Why it is worth fixing** — the documented use is producing an image for a bug report or a CI artifact, and every caller therefore writes `> file.svg`. That works in a shell and is awkward everywhere else: a `Makefile` loop, a PowerShell step on the Windows leg, or the `report` action, which renders several files in a row. It is also the one place where a redirect can silently produce an empty file when the command fails, because the shell truncates before the process runs.

**Fix** — add `--out PATH` to `render` in `crates/termlens-cli/src/main.rs`: write there instead of stdout, create nothing if the render fails, and keep stdout as the default so existing invocations are unchanged. Document it in `RENDER_USAGE` and `crates/termlens-cli/README.md`.

**Done when**
- [ ] `termlens render --svg --out x.svg a.snap` writes the same bytes stdout would have.
- [ ] A failing render (unreadable input) exits 2 and leaves no partial file behind.
- [ ] `crates/termlens-cli/tests/cli.rs` covers both.

---

*A good first contribution: one file, no architecture decisions. Start with
[CONTRIBUTING.md](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md)
— §1 lists every gate you can run locally, and §3 asks that a change land
with a test. Commits are Conventional Commits and need `git commit -s`
(DCO). Happy to review a draft PR early.*

PR #468, opened 2026-09-17
fix: reject empty render out path

### LINKED ISSUE #451: `termlens render --out=` fails with `termlens: : No such file or directory`, naming neither the flag nor the file
**Today** — `--out=` with nothing after the `=` produces a diagnostic that
names neither the flag nor the file:

```console
$ termlens render --text --out= screen.snap
termlens: : No such file or directory (os error 2)
```

It reads as a bug in termlens rather than as a mistake in the command. The
spelled-out form already gets this right:

```console
$ termlens render --text --out screen.snap
termlens: --out needs a PATH argument
```

**Why it is worth fixing** — `--out=` is the only `=`-spelled flag in the
tool whose empty value goes unnamed. Every other one refuses it by name:

```console
$ termlens inspect --cwd= echo hi
termlens: bad --cwd "", not an existing directory
$ termlens inspect --size= echo hi
termlens: bad --size "", expected e.g. 120x40
$ termlens diff --color= a b
termlens: --color takes auto, always or never, got ""
```

An empty value is what a shell hands over when a variable is unset —
`--out=$DEST` with `DEST` never assigned — so it is a realistic typo, and
the message is the one thing standing between the user and ten minutes of
looking for a missing input file.

**Fix** — in `render` in `crates/termlens-cli/src/main.rs`, the arm is

```rust
other if other.starts_with("--out=") => out_path = Some(&other["--out=".len()..]),
```

Treat an empty remainder as a missing argument and reuse the message the
other spelling already prints. Roughly one line, in one arm.

**Done when**
- [ ] `render --text --out= screen.snap` exits 2 and names `--out`.
- [ ] `render --text --out=real.txt screen.snap` still writes `real.txt`.
- [ ] `crates/termlens-cli/tests/cli.rs` pins the message, next to
      `render_out_writes_the_file_and_creates_none_when_the_render_fails`.

---

TayfurYldz:
Closes #451

## What changed
- reject an empty value in `render --out=` with the same named diagnostic as `--out` without a path
- add a CLI regression that pins exit code 2 and the exact message
- document the user-facing fix under Unreleased

## Why
The equals spelling accepted an empty slice as a path, so the later file error printed `termlens: : No such file or directory` instead of identifying `--out`.

## Testing
- baseline reproduced on current main: `render --text --out= ...` exits 2 with the unnamed file diagnostic
- `cargo test -p termlens-cli --test cli` (27 passed)
- `cargo clippy -p termlens-cli --all-targets -- -D warnings`
- `cargo test --workspace --all-features`
- `cargo fmt --all --check`
- `.github/scripts/check-dco.sh main..HEAD`
- `.github/scripts/check-no-ai-attribution.sh main..HEAD`
- `git diff --check`


vyncint (approved):
The fix itself is exactly right, and refusing the empty value rather than letting it reach the filesystem is the correct layer to catch it at. Verified on your branch:

```
render --text --out= before.snap   -> termlens: --out needs a PATH argument   (exit 2)
render --text --out=/dev/null …    -> exit 0, still writes
```

I pushed two adjustments on top:

**The test had landed between another test's doc comment and its function.** The `/// render --out exists because every caller was writing > file.svg … (#313)` comment sat directly above `render_out_writes_the_file_and_creates_none_when_the_render_fails`; inserting the new test under it silently re-pointed that rationale at the wrong function. Moved yours above it and gave it its own comment naming the unset-variable case, which is the one that motivated the issue.

**The CHANGELOG entry** needed the `### Fixed` heading (`[Unreleased]` was empty after 0.11.2 shipped) and the house phrasing — these read as a description of the behaviour rather than an imperative.

Also rebased twice, since #452 and #374 landed in the same section while this was open.

Thank you — and the exact-match assertion on stderr is the right call here; it would catch a stray second line.

PR #481, opened 2026-09-19
fix(cli): report a top-level flag as an unknown option

### LINKED ISSUE #475: A top-level unknown flag is reported as an unknown *command*, where every subcommand calls it an option
**Today** — a flag typed in place of a subcommand is reported as an unknown
*command*:

```console
$ termlens --verbose
termlens: unknown command "--verbose" (try --help)
$ termlens -v
termlens: unknown command "-v" (try --help)
```

Every subcommand calls the same kind of token an option:

```console
$ termlens render --pdf before.snap
termlens: unknown option "--pdf" (try `termlens render --help`)
```

The exit code is 2 in every case, which is right.

**Why it is worth fixing** — `-v` and `--verbose` are the two most likely
wrong guesses at a CLI, and both are flags, not commands. Being told
`--verbose` is not a *command* sends the reader looking for a subcommand
list when what they need is to know the flag does not exist — and the only
two top-level flags, `-h/--help` and `--version`, are exactly what a person
is reaching for when they type one of these.

It is the same diagnostic the subcommands already get right, so this is one
tool disagreeing with itself about what to call a `-`-prefixed token.

**Fix** — in `main()` in `crates/termlens-cli/src/main.rs`, the dispatch
ends with:

```rust
other => fail(&format!("unknown command {other:?} (try --help)")),
```

Split the arm: a token starting with `-` is an unknown option, anything
else stays an unknown command. Keep both at exit 2, and keep the `(try
--help)` hint — at the top level that is the right pointer, where a
subcommand points at its own `--help`.

Worth a thought while you are there: the top level accepts `--version` but
not `-V`, which is the conventional short form. Out of scope for this
issue, but say so in the PR if you have a view.

**Done when**
- [ ] `termlens --verbose` and `termlens -v` report an unknown **option**,
      still exit 2, and still point at `--help`.
- [ ] `termlens nonesuch` still reports an unknown **command** at exit 2
      (`check-cli-contract.sh` pins this one; run it).
- [ ] `crates/termlens-cli/tests/cli.rs` covers both shapes.

---

cestercian:
## Summary

A token starting with `-` in place of a subcommand was reported as an unknown *command*, so `termlens --verbose` / `termlens -v` pointed people at the subcommand list. Subcommands already call the same kind of token an option.

Top-level `-…` now says `unknown option` (exit 2, `try --help`). `termlens nonesuch` is unchanged.

Fixes #475

## Test plan

- [x] `cargo test -p termlens-cli --test cli` — 34 passed
- [x] `.github/scripts/check-cli-contract.sh` — pass
- [x] fmt + clippy clean


vyncint (approved):
Correct, and placed exactly where it has to be — after the three literal arms, before the catch-all, so `-h`/`--help`/`--version` still match first and only an *unrecognised* dash token is reclassified. The test pins the word as well as the exit, and keeps `nonesuch` a command, which is the half that could regress silently.

I pushed one trivial fix on top: the CHANGELOG bullet had a double blank line after it. Rebased onto `main` for #485 (an upstream `v1` tag move was reddening zizmor on all four of your PRs — nothing you did).

Verified locally: `check-cli-contract.sh` still passes, and the new test fails if the arm is removed.

Thank you.

PR #482, opened 2026-09-19
fix(cli): keep inspect --ansi stdout a saved screen

### LINKED ISSUE #478: `termlens inspect --ansi > file` is not a saved screen, though three documents promise without qualification that it is
**Today** — three documents promise, without qualification, that whatever
`inspect` writes to stdout is a saved screen. With `--ansi` it is not:

```console
$ termlens inspect --size 30x3 --ansi sh -c 'printf "\033[1;31mred\033[0m"' > shot.snap
$ termlens render --text shot.snap
termlens: shot.snap: could not parse a saved screen: line 2: '\u{1b}' is a
control character; the snapshot format never writes one
$ echo $?
2
$ termlens diff shot.snap shot.snap
termlens: shot.snap: could not parse a saved screen: line 2: …
```

Without `--ansi` the same command round-trips fine.

The three promises:

```
docs/STABILITY.md:164   Flags may be added; `inspect`'s stdout is a saved screen.
crates/termlens-cli/src/main.rs:16   the screen alone goes to stdout, so `inspect … >
                                     file` is a saved screen `diff` and `render` read back (#340)
INSPECT_USAGE           The screen goes to stdout and nothing else does, so `inspect … > file`
                        saves a screen that `termlens diff` and `termlens render` read back.
```

**Why it is worth fixing** — `--ansi` and `> file` are each reasonable, and
the combination is the natural thing to try when the plain rendering did
not show you what you needed and you want to keep it. The failure lands
later, on the `diff` or `render` that was the whole reason for saving, and
the message is about a control character rather than about the flag that
put one there.

The diagnostic itself is good — that is #376 doing its job. What is missing
is anything telling the user in advance, and a promise in `STABILITY.md`
that is true of every flag but one.

**Fix** — smallest honest version, and probably the right one: say it.
`--ansi` is a viewing mode, so `INSPECT_USAGE` gains a clause on the
`--ansi` line ("for reading, not for saving: the escapes it writes are not
the snapshot format"), the module doc's sentence gains the same
qualification, and `STABILITY.md`'s bullet says "`inspect`'s stdout is a
saved screen unless `--ansi` asks for a painted one".

There is a bigger option — have `--ansi` refuse to write to a
non-terminal, the way `--color auto` already decides by `is_terminal()` —
and an argument against it: a person legitimately pipes colour into `less
-R`. If you want to argue for it, say so in the PR; the documentation fix
stands either way and is the part that should not wait.

**Done when**
- [ ] `inspect --help` says the `--ansi` output is not a saved screen.
- [ ] The module doc and `docs/STABILITY.md` carry the same qualification.
- [ ] `check-candidate-statement.sh` and `check-readme-links.sh` pass.

---

cestercian:
## Summary

`inspect --ansi` still paints on a TTY. On a redirect or pipe it now writes a parseable saved screen (`Screen::with_styles`) instead of raw ANSI, matching the docs’ promise that inspect stdout is a saved screen.

Fixes #478

## Test plan

- [x] Workspace tests including redirect + TTY paths
- [x] clippy / fmt / DCO


vyncint (approved):
This is the harder of the two options #478 laid out, and the better one. Writing `with_styles` to a pipe keeps the colour the flag was asked for *and* keeps the file parseable — where documenting the loss, the cheap fix, would have left `inspect --ansi > file` producing something `render` refuses.

Two things I'd call out as done right. Updating `examples/inspect.rs` unprompted: it mirrors the command by design (#372) and has drifted twice before, so moving it in the same commit is exactly the habit that issue wanted. And testing *both* sides — `inspect_ansi_paints_on_a_terminal` driving it through a PTY so stdout really is a terminal, then asserting the cell's style rather than matching escape bytes — which is the stronger assertion and the one that survives a rendering change.

Verified on your branch: `inspect --ansi` to a pipe parses back through `render --text` where it exited 2 before, and carries `fg=1`/bold as a `styles:` block.

One follow-up, in the commit after yours: #484 lands the same idea for the plain path, and composed they made `is_terminal` decide everything — so a plain `inspect` at a prompt started printing a `styles:` block too. I split it on where stdout goes rather than on the flag, so a terminal keeps exactly what it printed before and every redirect is a saved screen. Your `--ansi` behaviour is unchanged either way; details on #484.

Rebased three times as the others landed, CHANGELOG-only each time. Thank you.

PR #483, opened 2026-09-19
fix(cli): treat render --out - as stdout

### LINKED ISSUE #469: `termlens render --out -` writes a file named `-` instead of to stdout, where every other operand treats `-` as a stream
**Today** — `--out -` is taken as a filename, so `render` creates a file
called `-` in the working directory:

```console
$ termlens render --text --out - before.snap
$ echo $?
0
$ ls -la ./-
-rw-rw-r-- 1 user user 140 Sep 19 18:34 ./-
```

Nothing was printed. The rendering is in a file whose name is the
conventional spelling of "standard output", and which most shells make
awkward to delete (`rm -` needs `rm ./-`).

**Why it is worth fixing** — `-` already means a stream everywhere else in
this tool. `load()` reads it as stdin for both `diff` operands and for
`render`'s, `STDIN_NAME` exists so a diagnostic can call it `<stdin>`, and
`RENDER_USAGE` says so in as many words: "`<a>` may be `-`, meaning
standard input." One flag out of the set treats it as an ordinary path, and
the failure is silent — the exit code is 0 and the screen went somewhere
the user did not ask for.

It also breaks the shape a pipeline wants: `termlens render --svg --out -
in.snap | some-filter` is the natural way to write "render to a stream",
and it currently writes a file and pipes nothing.

**Fix** — in `render` in `crates/termlens-cli/src/main.rs`, `out_path` is
consumed at the bottom:

```rust
match out_path {
    Some(path) => match std::fs::write(path, &out) { … },
    None => print(&out),
}
```

Treat `Some("-")` the way `None` is treated — `print(&out)`, which already
handles a closed reader. Both spellings (`--out -` and `--out=-`) arrive
here as the same string, so one place covers them. Say it in
`RENDER_USAGE` next to the sentence about `<a>`, and mind that a real file
named `-` can still be addressed as `./-`, which is the usual escape hatch
and worth a word.

**Done when**
- [ ] `render --text --out - before.snap` prints to stdout and creates no
      file, in both spellings.
- [ ] `--out ./-` still writes a file named `-`, for anyone who meant it.
- [ ] A named path still writes a file, and a failing render still leaves
      none (#313).
- [ ] `crates/termlens-cli/tests/cli.rs` covers it, next to
      `render_out_writes_the_file_and_creates_none_when_the_render_fails`.

---

cestercian:
## Summary

`termlens render --out -` now writes to stdout (like every other `-` operand) instead of creating a file named `-`. `--out ./-` still writes a file named `-`.

Fixes #469

## Test plan

- [x] `cargo test -p termlens-cli --test cli` — 34 passed
- [x] `.github/scripts/check-cli-contract.sh` — pass
- [x] fmt + clippy clean


vyncint (approved):
Right fix and right shape. Reusing the existing dash constant means both spellings (`--out -` and `--out=-`) arrive at the same branch for free, and `Some(STDIN) | None => print(&out)` reuses the broken-pipe handling `print` already had — so `render --out - | head` is a clean exit rather than a panic.

The test is the part I'd single out: running in a temp `current_dir` so a regression that writes `./-` shows up there rather than in the suite root, and still asserting `--out ./-` writes a real file named `-`. That last case is what stops the fix from being over-broad.

I pushed one small thing: the `STDIN` constant's doc said "the operand that means standard input", which is now only half true — it is the dash in either position. Broadened the comment rather than renaming, since the name is load-bearing in three other places.

Verified: `render --text --out - before.snap` prints and leaves no file; `check-cli-contract.sh` passes. Rebased twice — for #485 and again after #481 — both CHANGELOG-only.

Thank you.

PR #484, opened 2026-09-19
fix(cli): keep styles when inspect redirects to a file

### LINKED ISSUE #454: `termlens inspect > file` strips every style, so `termlens diff` between a red screen and a green one prints "no difference"
**Today** — `inspect` prints the plain rendering, so a screen saved with
`> file` has no `styles:` block, and every colour, bold and reverse-video
run is gone. `diff` then cannot see a colour change at all:

```console
$ termlens inspect --size 20x2 sh -c 'printf "\033[1;31mred\033[0m"' > red.snap
$ termlens inspect --size 20x2 sh -c 'printf "\033[1;32mred\033[0m"' > green.snap
$ cat red.snap
size: 20x2  cursor: 0,3
red

$ termlens diff red.snap green.snap
no difference
$ echo $?
0
```

Bold red against bold green, and `diff` says the screens are the same
picture and exits 0.

**Why it is worth fixing** — `inspect … > file` is promised as the way to
produce a saved screen: `termlens --help`, the README, `SKILL.md` §9b and
the module doc all say the file `diff` and `render` read back is what
`inspect` writes to stdout ([#340](https://github.com/vyncint/termlens/issues/340)).
It does read back — with a quarter of the screen's content silently
missing. `render --text` on the same screen keeps the block, so the tool's
own two text renderings disagree about what a saved screen is, and the one
users are pointed at is the lossy one.

`docs/DESIGN.md` §3 is explicit that the plain rendering is style-blind and
that `with_styles()` "is the rendering to compare for *looks the same*" —
which is exactly what `diff` is for. Nowhere in the CLI's own help does it
say the file it just wrote is style-blind.

**Fix** — in `inspect_render` in `crates/termlens-cli/src/main.rs`, the
non-`--ansi` branch is `screen.to_string()`. `render --text` a few lines up
uses `screen.with_styles()`. Making `inspect` write the same rendering is a
one-line change and makes the saved file lossless.

The trade-off to state in the PR: it adds a `styles:` block to what a human
sees at a prompt, which is noise when they only wanted to look. If that is
judged the wrong default, the alternative is to keep the output as it is
and say so plainly in `INSPECT_USAGE`, the README and `SKILL.md` — "the
saved screen carries no styles; use `render --text` for one that does".
Either way the current state, where the promise and the behaviour disagree
in silence, should not survive.

Worth checking while you are there: `strip_inspect_trailer` looks at the
last non-blank line, and a `styles:` block changes what that line is.

**Done when**
- [ ] `termlens diff` between the red and green screens above exits 1 and
      names the style change — or `inspect --help` says the file has no
      styles, and the README and `SKILL.md` say it too.
- [ ] A screen saved by `inspect` still parses, still round-trips through
      `render --text` and `render --json`, and a file saved by an older
      `inspect` (trailer included) still reads.
- [ ] `crates/termlens-cli/tests/cli.rs` pins whichever behaviour is
      chosen, and the CHANGELOG records it.

---

cestercian:
## Summary

`inspect > file` now writes a `with_styles` snapshot so colour-only differences show up in `diff`.

Fixes #454

## Test plan

- [x] CLI tests including red vs green diff
- [x] clippy + cli-contract


vyncint (approved):
The diagnosis is right and the test is the best of the four: it does not stop at "the block is there" but parses both screens back, asserts `fg=1` against `fg=2` on the cell, runs `diff` and checks it *names* the run, then round-trips through `render --json` — and finally reads a file with a 0.10-style trailer appended, which is the backward-compatibility case nobody would think to write.

One change on top, and it is a design call rather than a defect in your work.

Your PR and #482 both put a decision in `inspect_render`, and composed they made `is_terminal` decide everything: `--ansi` painted on a terminal, and *everything else* — including a plain `inspect` at a prompt — got `with_styles`. So every `termlens inspect` started printing a `styles:` block at a person who only wanted to look. #454 weighed exactly that noise and called it the price of the simple fix; once #482 put `is_terminal` in this function the price stopped being necessary.

So I split it on where stdout goes rather than on the flag:

```
terminal + plain   -> to_string        (exactly as before)
terminal + --ansi  -> to_ansi          (#482, unchanged)
pipe/redirect      -> with_styles      (your fix, and #478's)
```

That is the title you filed under — *keep styles when inspect **redirects** to a file* — and it is what your CHANGELOG bullet describes. Worth knowing why it needed saying: **every test in this file reads `inspect` through a pipe**, yours included, so all of them pass either way and nothing would have caught the terminal turning noisy. `inspect_on_a_terminal_stays_plain` is there so the next simplification back to one branch goes red.

Also folded in: `INSPECT_USAGE` states the split once for both flags rather than only under `--ansi`, and `examples/inspect.rs` takes the same shape as the command it mirrors.

Verified: `diff` between a bold-red and a bold-green screen now exits 1 and names the run — the original #454 reproduction — and `check-cli-contract.sh` passes.

Thank you — four issues in a day, and each of them read the issue rather than just the title.

Issue #480, opened 2026-09-19
Nothing checks that `examples/inspect.rs` still mirrors `termlens inspect`, and it has drifted twice


vyncint:
**Today** — `crates/termlens/examples/inspect.rs` is kept deliberately in
step with `termlens inspect`. Its own comments say so:

```rust
// The screen to stdout, the trailer to stderr (#340): … `termlens inspect`
// splits the same two streams the same way.
```

Nothing checks it. The two files carry separate copies of the flag parser,
the usage text, the `REAP_GRACE` constant and the wait, and only a reviewer
noticing keeps them equal:

```console
$ grep -rn 'examples/inspect' .github/scripts/*.sh .github/workflows/*.yml tools/ CONTRIBUTING.md
$ echo $?
1
```

**Why it is worth fixing** — it has drifted, twice, and both times the
repair was its own commit:

```console
$ git log --oneline -3 -- crates/termlens/examples/inspect.rs
4306c98 fix(cli): inspect resolves on exit or idle, whichever comes first (#465)
53f9fda feat(cli): accept `--flag=value` on inspect (#437)
51bad7b fix(example): align the inspect example with the CLI contract (#443)
```

#443 exists only because the example had fallen behind. In #465 it fell
behind again — the command's wait changed and the example kept the old one,
still describing it in its own `--help`, and that was caught in review
rather than by anything automatic. The two texts differ today in a way
nobody chose: the example documents `--` and the command does not
([#453](https://github.com/vyncint/termlens/issues/453)), while the command
explains `--ansi` and the example does not.

This is the shape [#386](https://github.com/vyncint/termlens/issues/386)
had — three hand-maintained copies of one version, two of which went stale
twice — and the answer there was `check-report-pin.sh`.

**Fix** — the cheap, useful version is a test rather than a script. In
`crates/termlens/tests/inspect.rs`, which already builds and drives the
example, run the example and `termlens inspect` against the *same* program
and assert they agree on what matters:

- the same trailer for a program that exits, and for one that outlives the
  wait;
- the same screen on stdout;
- the same exit codes for a bad flag and a missing program.

That pins the contract without freezing the prose — the two usage texts
legitimately differ (`usage: inspect` vs `usage: termlens inspect`), so
comparing them byte for byte is the wrong test.

Pick a program whose behaviour is not timing-sensitive, and mind
CONTRIBUTING §3 on where a short deadline belongs.

**Done when**
- [ ] A test drives both binaries on the same program and asserts they
[...]

PR #486, opened 2026-09-20
Nothing checks that examples/inspect.rs still mirrors termlens inspect, and it has drifted twice  #480

bernalalexis-try:
I wrote a test for #480 in crates/termlens/tests/inspect.rs that runs the example and termlens inspect on the same program and compares screen, trailer and exit codes. It fails if #465 is reverted in either one.


vyncint (commented):
Thank you for this — the tests are the right answer to #480, and better than what the issue sketched. Comparing the *contract* (screen, trailer, exit code) while deliberately not comparing the usage prose is the correct line, and saying so in a comment means the next person will not 'fix' it by diffing the help text. Running both children under `thread::scope` so a pair costs the slower rather than the sum is a nice touch, and the four still-running / deadline / EOF cases cover exactly the paths that drifted in #443 and #465.

I verified it locally: the six new parity tests pass (12 in `tests/inspect.rs` now), and the full suite is 48 result lines with no failures.

**I pushed one commit to fix what was reddening CI**, which was not the tests:

The branch turned **all eighteen shell scripts from 100755 to 100644** — every gate under `.github/scripts/` and every selftest under `tools/`. CI runs them as `./check-….sh`, so nine checks failed at once on a permission error, which is why the failures looked unconnected to what you were adding. Usually a checkout on a filesystem that does not carry the executable bit. Restored to match `main`, and with that the suite, the gates and the selftests all pass.

I also removed two files that came along by accident: `PR_DESCRIPTION.md` (that text belongs in the PR body) and `inspect-parity.patch` (a copy of this branch's own diff). Your tests are untouched.

**One thing left, and it has to be you.** `commit-policy` is still red:

```
Commit c95c7b7 has no Signed-off-by matching its author <bernalalexis030@gmail.com>.
```

The DCO sign-off is a certification about the origin of the code, so a maintainer cannot add it on your behalf — and mechanically it would not help either: `check-dco.sh` requires the sign-off email to match the *author* email, so my name on your commit fails the same check. From your branch:

```sh
git pull                       # picks up the permissions fix
git rebase --signoff origin/main
git push --force-with-lease
```

That signs both of your commits as you and is the only thing between this and a merge. Everything else is green or on its way. Ping me when it is pushed and I will merge it.

bernalalexis-try:
Resolved

vyncint (approved):

Issue #470, opened 2026-09-19
The AI-attribution gate is the policy this repo enforces hardest, and the only way to know it still fails is to run seven commands by hand


vyncint:
**Today** — `.github/scripts/check-no-ai-attribution.sh` carries its own
test plan as a comment in its header:

```sh
# tests: exercise locally on a scratch branch before trusting it in CI —
#   git checkout -b scratch/attribution
#   git commit --allow-empty -s -m "test: clean commit"
#   git commit --allow-empty -s -m "test: bad commit" \
#     -m "Co-Authored-By: Example Bot <example[bot]@users.noreply.github.com>"
#   .github/scripts/check-no-ai-attribution.sh main..HEAD  # must fail once
#   …
```

Seven commands, run by hand, by someone who remembers to. Nothing in CI
runs them, so the only evidence the gate can still fail is a human who
tried it that day.

**Why it is worth fixing** — three *lesser* gates already have one:

```
tools/semver-gate-selftest/run.sh
tools/link-gate-selftest/run.sh
tools/report-pin-gate-selftest/run.sh
```

CONTRIBUTING §1 lists all three, and their whole argument is that a gate
which cannot be shown to fail is a gate nobody can trust. This is the
policy the repository enforces hardest — `CLAUDE.md` and `AGENTS.md` both
name it, and it is the one rule that cannot be repaired after the fact,
because a bad trailer on `main` is in the history for good.

The gate is also the most intricate of the set: a vendor list, a robot
emoji, author *and* committer identities, a Dependabot carve-out that
applies to identity but not to the message, and a `Co-authored-by` GitHub
itself appends on squash. Each of those is a branch that could stop
matching without anything going red.

**Fix** — `tools/attribution-gate-selftest/run.sh`, modelled on
`tools/report-pin-gate-selftest/run.sh` (61 lines, the smallest of the
three). The cases are already written in the header comment; the work is
turning them into a script that builds a scratch repository, runs the gate
and asserts pass or fail per case. Worth covering at least:

- a clean signed commit passes;
- an AI co-author trailer fails;
- a "Generated with" watermark and the robot emoji each fail;
- the Dependabot identity carve-out passes on identity but still fails on
  a watermark in its message;
- an unresolvable range and an empty range both **fail** rather than
  reporting OK on nothing (this is termlens#214, and it is the failure
  mode a selftest exists to catch).

Build the scratch history in a `mktemp -d`, not in the checkout. Add the
[...]

Issue #471, opened 2026-09-19
The DCO gate's "can it fail?" is a hand-run recipe too, and its squash-merge exemption is the part nothing checks


vyncint:
**Today** — `.github/scripts/check-dco.sh` has the same shape as its
neighbour: a hand-run recipe in a comment, and nothing that runs it.

```sh
# tests: exercise locally on a scratch branch before trusting it in CI —
#   git checkout -b scratch/dco
#   git commit --allow-empty -m "test: signed"  -s
#   git commit --allow-empty -m "test: unsigned"
#   .github/scripts/check-dco.sh main..HEAD   # must fail on the 2nd commit
#   # and the composed-squash path, which must pass only when named:
#   …
```

**Why it is worth fixing** — the interesting half is the second one. The
gate takes an optional second argument naming the one commit GitHub may
have composed itself when squash-merging, and that commit is exempted from
the sign-off rule because GitHub authors it as `noreply@github.com`. So the
gate has a deliberate hole, and its correctness is entirely about the hole
being exactly one commit wide and only when named. Nothing checks that:

```sh
.github/scripts/check-dco.sh main..HEAD            # must fail
.github/scripts/check-dco.sh main..HEAD "$(git rev-parse HEAD)"  # passes
```

An exemption that silently widened would let unsigned commits onto `main`
and look green doing it. The same unresolvable-range case as its sibling
matters too: a range git cannot read must fail rather than report OK on
commits it never read (termlens#214).

**Fix** — `tools/dco-gate-selftest/run.sh`, modelled on
`tools/report-pin-gate-selftest/run.sh`. Cases: a signed commit passes; an
unsigned one fails; a sign-off whose email differs from the author's fails;
the composed-squash commit fails unless named and passes when named; a
merge commit is exempt; an unresolvable range and an empty range both fail.
Build the history in a `mktemp -d`. Add the line to CONTRIBUTING §1 beside
its siblings, or `check-ci-gates-listed.sh` will say so.

This is a sibling of the attribution-gate selftest; either can land first
and the second should follow whatever shape the first sets.

**Done when**
- [ ] `tools/dco-gate-selftest/run.sh` passes and covers every case above.
- [ ] Widening the composed-squash exemption on purpose makes it fail.
- [ ] CONTRIBUTING §1 lists it and `check-ci-gates-listed.sh` is green.

---

*A good first contribution: one file, no architecture decisions. Start with
[...]

PR #487, opened 2026-09-20
test(ci): add AI attribution gate self-test (#470)

hs150:
Add tools/attribution-gate-selftest/run.sh modelled on the sibling report-pin self-test. Exercises each pattern and failure mode in check-no-ai-attribution.sh against an isolated scratch repository: clean signed commit, AI co-author trailer, tool watermarks and robot emoji, bot author identity, Dependabot carve-outs (identity and squash trailer), Dependabot carrying message watermarks, and unresolvable and empty commit ranges.

Wire the self-test into CONTRIBUTING.md §1 and .github/workflows/ci.yml, and update check-no-ai-attribution.sh header to reference the runner.

## Checklist

- [ ] Linked an issue (or explained above why none exists)
- [ ] Tests added/updated for the change
- [ ] **All commits are signed off** (`git commit -s`) — see [CONTRIBUTING.md §5](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md#5-developer-certificate-of-origin-dco)
- [ ] **No AI attribution trailers** (no AI co-authors, "Generated with" footers, or bot identities) — see [CONTRIBUTING.md §6](https://github.com/vyncint/termlens/blob/main/CONTRIBUTING.md#6-ai-tooling-policy)
- [ ] Snapshot changes (if any) were reviewed with `cargo insta review`, not blind-accepted


vyncint (approved):
This does the whole of #470 and does it the way the other three selftests do. Ten cases, a scratch repository in `mktemp -d` rather than the checkout, and the `expect` helper reporting every case before exiting non-zero — so one broken rule does not hide the rest.

I checked the thing that actually matters for a selftest, which is whether it fails when the gate does. Two mutations against your branch:

```
drop the robot-emoji rule        -> FAIL  a robot emoji watermark fails   exit 0, expected nonzero
widen the carve-out to all bots  -> FAIL  a bot author identity fails     exit 0, expected nonzero
```

Both caught, each naming the case that stopped holding. That second one is the one I care about most: the Dependabot exemption is the deliberate hole in this gate, and case 7 (the squash-merge `Co-authored-by` GitHub composes itself) plus case 8 (Dependabot still failing on a watermark) pin its two edges. Those are the cases someone writing this from the header comment alone would most likely skip.

Replacing the hand-run recipe with a pointer rather than leaving both is right — two copies of a test plan drift, which is the problem this repository keeps solving. And thank you for the CONTRIBUTING line; `check-ci-gates-listed.sh` would have caught its absence, but you did not make it.

Nothing to change. #471 is the same shape for the DCO gate if you want it — its composed-squash exemption is the interesting half, much like the carve-out here.

